PR #33, opened 2018-12-30
Add 'inject' command

### LINKED ISSUE #31: Add new workspace:inject --kubeconfig command
The command syntax:

```
Inject configurations and tokens in a Che Workspace

USAGE
  $ chectl workspace:inject

OPTIONS
  -k, --kubeconfig                  Inject the local Kubernetes configuration
  -h, --help                        show CLI help
  -w, --workspace=workspaceid       Target workspace
  -c, --container=containerid      [default: dev] Target container
  -n, --chenamespace=chenamespace   [default: kube-che] Kubernetes namespace where the workspace is deployed
```

When the command is executed`chectl` should look at local `~/.kube/config` and:
- generate a new ~/.kube/config to copy it in the target container
- copy certificates and keys in the target container

More details about the expected behaviour of the command:

- If a `~/.kube/config` already exist in the target container the command should fail (we may later add a `--force` flag but that's not in the scope of this issue)
- The `--workspace` flag can be omitted only if there is just one running workspace (if will fail otherwise)
- If the `--container` flag is omitted the default container should be the che-dev (it will fail if che-dev doesn't exist)

azatsarynnyy:

closes #31 

The PR introduces `chectl workspace:inject` command for pushing the local Kubernetes configuration into a Che Workspace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ USAGE
 * [`chectl server:start`](#chectl-serverstart)
 * [`chectl server:stop`](#chectl-serverstop)
 * [`chectl server:update`](#chectl-serverupdate)
+* [`chectl workspace:inject`](#chectl-workspaceinject)
 * [`chectl workspace:list`](#chectl-workspacelist)
 * [`chectl workspace:start`](#chectl-workspacestart)
 * [`chectl workspace:stop`](#chectl-workspacestop)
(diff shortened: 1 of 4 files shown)


azatsarynnyy:
@l0rd could you take a look
Note, for now the copied information is limited with `minikube` context only since I'm not sure whether it makes sence to inject the entire local kubeconfig.

PR #70, opened 2019-03-28
Only start che on minishift and minikube if userland-proxy is disabled

benoitf:
Ensure userland-proxy has been defined on minikube or minishift before deploying che.

- [X] Also add a new script to auto-fix the code format failures

Change-Id: Iea000a65c26cca00972deea498d6131eb3c5709b

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
   "scripts": {
     "postpack": "rm -f oclif.manifest.json",
     "posttest": "tslint -p test -t stylish",
+    "tslint-fix": "tslint --fix -p test -t stylish",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -100,6 +100,7 @@
   "scripts": {
     "postpack": "rm -f oclif.manifest.json",
     "posttest": "tslint -p test -t stylish",
+    "tslint-fix": "tslint --fix -p test -t stylish",
```
l0rd: @benoitf what is this for?

benoitf: It's to auto-format

Issue #71, opened 2019-03-28
server:stop does not work on minishift


jeffmaury:
Seems it is bound to minikube and helm

l0rd:
I am working on it

l0rd:
Fixed by #76

Issue #56, opened 2019-03-13
server:stop should be renamed server:delete and it should delete all resources


l0rd:
I usually use the following script to delete all Che resources. The command `server:delete` should do the same.

```bash
#!/bin/bash

namespace="che"

k delete deployment --all -n ${namespace}; \
k delete service --all -n ${namespace}; \
k delete ingress --all -n ${namespace}; \
k delete configmap che-operator -n ${namespace}; \
k delete rolebinding che che-operator che-workspace-exec che-workspace-view -n ${namespace}; \
k delete serviceaccount che che-operator che-workspace -n ${namespace}; \
k delete pod/che-operator -n ${namespace}; \
k delete pvc/postgres-data -n ${namespace}; \
k delete pvc/che-data-volume -n ${namespace}; \
k delete configmap che -n ${namespace}; \
oc delete route --all -n ${namespace}
````

l0rd:
Fixed by #76

PR #76, opened 2019-04-01
Chang server:stop behavior and added server:delete

l0rd:
With this PR the command `server:stop`:
-  calls Che server shutdown endpoint (`/api/system/stop?shutdown=true`) 
- scales down the deployment to zero (as well as keycloak and postgres if needed)

In this PR is also included a new command: `server:delete`. It rapidly purges all Che resources (deployments, pods, services, routes, cm, pvc etc...) to start from scratch.

Some other changes included in this PR:

- Automatically find out if it's Kubernetes or OpenShift (currently only stop and delete commands)
- server:stop is the first command that supports authenticated Che API (user has to provide the access token as a param, I would like to create a new auth:token command that would help with that)
- Added 2 end2end tests for minikube and minishift: I would like to setup a circleCi job that use them for PR validation
- Added an option to use verbose Listr renderer for all commands because it's better for automated tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ USAGE
 <!-- commands -->
 * [`chectl autocomplete [SHELL]`](#chectl-autocomplete-shell)
 * [`chectl help [COMMAND]`](#chectl-help-command)
+* [`chectl server:delete`](#chectl-serverdelete)
 * [`chectl server:start`](#chectl-serverstart)
 * [`chectl server:stop`](#chectl-serverstop)
 * [`chectl server:update`](#chectl-serverupdate)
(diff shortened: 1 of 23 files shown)


l0rd:
Fixes #71 and #56

benoitf:
Hello,

I tried on minikube

`server:start` then `server:stop` and now `server:start` but it's stuck at

```
  ❯ ✅  Post installation checklist
    ❯ Che pod bootstrap
      ⠦ scheduling
        downloading images
```

benoitf:
` kubectl get all --namespace che`

```
$ kubectl get all --namespace che
NAME               TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)             AGE
service/che-host   ClusterIP   10.103.228.218   <none>        8080/TCP,8087/TCP   15m

NAME                  READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/che   0/0     0            0           15m

NAME                             DESIRED   CURRENT   READY   AGE
replicaset.apps/che-5959c547b6   0         0         0       15m

```

l0rd:
@benoitf you are right. I had not made the changes needed to support this scenario in this PR: `server:start` never check if che server was already deployed or not.

The only way to restart Che is `kubectl scale deployment che --replicas=1`. But that doesn't make this PR consistent so I will try to look how much changes are needed to support your scenario. If it needs too much work I would rather do it in another PR.

benoitf:
ok, I was not able to test on minishift as each time it pull all images it takes me at least one hour :-)

l0rd:
I am going to merge this PR since fixing `server:start` properly doesn't look so easy. I have opened #77 to track that.

Issue #72, opened 2019-03-29
Notify user when there is a new version of chectl or allow to update


benoitf:
- https://oclif.io/docs/releasing.html#autoupdater
- https://github.com/oclif/plugin-update
- https://www.npmjs.com/package/@oclif/plugin-warn-if-update-available

nickboldt:
Having discussed the options and requirements with @l0rd, here's the plan:

[ ] take travis script, port to jenkinsfile
[ ] set up job in crw jenkins to run jenkinsfile on any git commit events
[ ] switch build from pkg to npx oclif-dev pack
[ ] push resulting tarball via rsync over ssh to https://dl.jb.o/jbosstools/che-incubator/chectl/dist/chectl.tar.gz

benoitf:
I'm not sure why pkg shouldn't be used ? As it's offering single binary without pre-requesites and how it is related to changing CI as well

l0rd:
@nickboldt wants to help us setting up a Jenkins job that would publish the artifacts to download.jboss.org. 

He needed more details to setup the Jenkins build and I've pointed him to the travis.yml and told him that we need to move from `pkg` to `oclif-dev pack` to be able to use the `autoupdate` feature. At least that's what I understand from docs [here](https://oclif.io/docs/releasing#autoupdater) and [here](https://oclif.io/docs/faqs#i-want-a-single-binary-cli-like-with-go).

And [I wanted to give a try to the tarball packaging](https://github.com/che-incubator/chectl/issues/36#issuecomment-457013341) in any case because:
- `pkg` has a couple of annoying issues
- oclif documentation suggest not to use pkg

l0rd:
@nickboldt I have run ` npx oclif-dev pack` locally and the result is:

```
./dist/
|-- chectl-v0.0.2
    |-- chectl-v0.0.2-darwin-x64.tar.gz
    |-- chectl-v0.0.2-linux-arm.tar.gz
    |-- chectl-v0.0.2-linux-x64.tar.gz
    |-- chectl-v0.0.2-win32-x64.tar.gz
    |-- chectl-v0.0.2-win32-x86.tar.gz
    `-- chectl-v0.0.2.tar.gz
 ```

so the format of the URLs will be `https://dl.jb.o/jbosstools/che-incubator/chectl/dist/chectl-vX.Y.Z/chectl-vX.Y.Z-<os>-<arch>.tar.gz`

nickboldt:
[...]

PR #79, opened 2019-04-03
#72 - Jenkinsfile to build chectl and publish to dl.jboss.org

nickboldt:
Signed-off-by:...

#72 - initial commit

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ OPTIONS
   --all  see all commands in CLI
 ```
 
-_See code: [@oclif/plugin-help](https://github.com/oclif/plugin-help/blob/v2.1.4/src/commands/help.ts)_
+_See code: [@oclif/plugin-help](https://github.com/oclif/plugin-help/blob/v2.1.6/src/commands/help.ts)_
 
 ## `chectl server:delete`
 
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ node_modules
 chectl-v*.tgz
 coverage
 .vscode
-
+bin/chectl-*
+tsconfig.tsbuildinfo
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,50 @@
+#!/usr/bin/env groovy
+
+// PARAMETERS for this pipeline:
+// branchToBuildCTL = refs/tags/20190401211444 or master
+// DESTINATION = user@host_or_ip:/path/to/che-incubator/chectl/
+
+def installNPM(){
+	def nodeHome = tool 'nodejs-10.9.0'
+	env.PATH="${env.PATH}:${nodeHome}/bin"
+	sh "npm install -g yarn pkg"
+	sh "npm version"
+}
+
+def CTL_path = "chectl"
+//def VER_CTL = "VER_CTL"
+def SHA_CTL = "SHA_CTL"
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -91,7 +91,7 @@ OPTIONS
   --all  see all commands in CLI
 ```
 
-_See code: [@oclif/plugin-help](https://github.com/oclif/plugin-help/blob/v2.1.4/src/commands/help.ts)_
```
l0rd: This file (README.md) is generated automatically (with `yarn pack`). Hence what you probably want to do is:
- bump `@oclif/plugin-help` to the new version (old version is [here](https://github.com/che-incubator/chectl/blob/master/yarn.lock#L286)): `yarn upgrade @oclif/plugin-help`
- update README.md: `yarn pack`
- update your PR with the new `yarn.lock` and the new `README.md`

nickboldt: you mean https://github.com/che-incubator/chectl/blob/1a37091160503e96b380625f5532c72c517f09be/yarn.lock#L286-L288 ?

l0rd: Yes:`yarn.lock` is another file you don't update manually. You need to use the command `yarn upgrade @oclif/plugin-help` instead.

nickboldt: ok, I've run the above commands and updated this PR to the resulting changes. Also updated yarn since I was running with an old version.

package.json
```diff
@@ -34,7 +34,8 @@
     "mkdirp": "^0.5.1",
     "ncp": "^2.0.0",
     "node-notifier": "^5.3.0",
-    "tslib": "^1"
+    "tslib": "^1",
+    "yarn": "^1.15.2"
```
l0rd: You have added `yarn` as a dependency of the project here. I suspect that in the process of updating `yarn` on your laptop you have added it as a chectl dependency as well (i.e. you have run `yarn add yarn`). If that's the case you should revert it (using command `yarn remove yarn`) and update your commit.

nickboldt: I didn't add it explicitly but given the build DOES require yarn why is it bad to state it here?

l0rd: Do you usually add maven in your pom.xml?

nickboldt: no, just maven plugins... so if here you're saying it's not valuable to define a minimum version of yarn that's been tested to work here, I'll remove it.

benoitf:
hello, what is the status of this PR ?

nickboldt:
It awaits feedback from Mario and Florent. 

It runs here: 

https://codeready-workspaces-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/CRW_CI/view/Builds/job/eclipse-chectl_master/ 

And you can test the output here:

http://download.jboss.org/jbosstools/che-incubator/chectl/dist/

package.json
l0rd: @nickboldt please remove `yarn` as a dependency so that we can proceed merging this PR

nickboldt: I think I already did last week, but missed a reference in yarn.lock. Force pushed an update to this PR. Is that better?

nickboldt:
"ERROR: Permission to che-incubator/chectl.git denied to nickboldt."

l0rd:
@nickboldt invitation sent. You should become a committer once you accept it.

PR #84, opened 2019-04-17
Add command devfile:generate

l0rd:
Add command `devfile:generate` as described in #81:

```
USAGE
  $ chectl devfile:generate

OPTIONS
  -h, --help             show CLI help
  --command=command      command to include in the workspace
  --language=language    add support for a particular language. Currently supported languages: java,typescript,go,python,c#
  --namespace=namespace  Kubernetes namespace where the resources are defined
  --plugin=plugin        Che plugin to include in the workspace
  --project=project      source code project to include in the workspace
  --selector=selector    label selector to filter the Kubernetes resources
``` 

Example of usage:

```bash
chectl devfile:generate \
           --selector="app=guestbook" \
           --language=java \
           --project='{"name": "guestbook", "source": "https://github.com/kubernetes/examples.git"}'
```


l0rd:
@sleshchenko it's still WIP: I would like to do some better type checking + better error messages is something goes wrong.

It currently filters out only Deployments when `--selector` flag is specified. What other objects you think we should filter out? I don't think that we should filter out Pods because you may end with 3 pods if you have a deployment with replica=3.

And the result is of kind `DeploymentList`. Do we support it or should we transform it in a `List`?

sleshchenko:
@l0rd I think we should filter also:
- **Services**: they may be used for communications between containers;
- **PVC**: they are used for data persisting, typical usage DB data;
- Ingress/Route: not sure if we really need it, but it would allow a user to access his application without `port discovery plugin` (not sure that it's the right name for it)
- **ConfigMap**: typically is used by an application and Deployment will fail if it is absent;
- Secret: Not sure if it is a good idea to export secrets but user deployment may fail without it;
- Pod: there may be cases for bare pods, but it's not critical now.

>  I don't think that we should filter out Pods because you may end with 3 pods if you have a deployment with replica=3.
Pods created with deployment have the corresponding label, so you can filter them out.

> And the result is of kind DeploymentList. Do we support it or should we transform it in a List?

I think `DeploymentList` should work as well, but I think that deployments are not enough, we need at least services + pvcs.

l0rd:
@sleshchenko makes sense I will return a `List` with all this stuff

src/commands/devfile/generate.ts
```diff
@@ -0,0 +1,266 @@
...
+import { Devfile, DevfileComponent, DevfileProject, ProjectSource, TheEndpointName } from '../../api/devfile'
+import { KubeHelper } from '../../api/kube'
+
+const kube = new KubeHelper()
+const stringLitArray = <L extends string>(arr: L[]) => arr
+const languages = stringLitArray(['java', 'typescript', 'go', 'python', 'c#'])
+export type Language = (typeof languages)[number]
+
+const LanguagesComponents = new Map<Language, DevfileComponent>([
+  ['java', {type: TheEndpointName.ChePlugin, alias: 'java-ls', id: 'org.eclipse.che.vscode-redhat.java:0.38.0'}],
```
garagatyi: @l0rd when are you going to merge this PR? If it lands after the new plugin notation https://github.com/eclipse/che/pull/13204 we'll need to update this PR

l0rd: @garagatyi hopefully by the beginning of the week. Anyway thanks for the heads up. I don't know what will be merged first but it should be pretty straightforward to fix but we need to think about it.

l0rd: I will work on this PR today and will update it with the new naming convention

garagatyi: ok

sunix:
@l0rd any update on this PR ? i was looking for it on the master branch :)

sunix:
@l0rd FYI 

:warning: i have rebased and added few this branch :warning: 

I have added few option to this:
- the ability to set the workspace name
    ```
    chectl devfile:generate --name=myworkspace
    ```
- the ability to use another editor. at the moment choosing from theia-next or theia-1.0.0. If not set, default would be theia-1.0.0 but i wanted to use next.
    ```
    chectl --editor=theia-next
    ```
- the ability to add a dockerimage component:
    ```
    chectl --dockerimage=jekyll/builder
    ```
    would generate
    ```yaml
    components:
      - alias: jekyll-builder
        type: dockerimage
        image: jekyll/builder
        memoryLimit: 512M
        mountSources: true
        command:
          - tail
        args:
          - '-f'
          - /dev/null
        ```
- simplify 'command': previously we had to provide a json of the command object. Now we just pass the command. Note that it will try to use the first component it founds and the first project it founds
    ```
    chectl devfile:generate  --dockerimage=jekyll/builder --command="mkdir -p /projects/blog
.sunix.org.git/_site && jekyll serve --host 0.0.0.0"
    ```

sleshchenko:
This PR becomes bigger and bigger. I would proceed with merging some part of arguments for `devfile:generate` command and then implement/discuss other arguments in other PRs/issues.
I think it makes sense to merge at least that part that we demonstrated on DockerCon
```
chectl devfile:generate \
      --namespace='nodejs-app' \
      --selector='app.kubernetes.io/name=employee-manager' \
      --git-repo='https://github.com/sleshchenko/NodeJS-Sample-App.git' \
      --language=typescript > generated.devfile.yaml
```

@sunix

> `chectl --editor=theia-next`

I think that next here is a version but not a name of editor, and it is better to use some separator like `/` or `:` to separate name and version. In plugin registry `/` is used for such purposes.

> chectl devfile:generate  --dockerimage=jekyll/builder --command="mkdir -p /projects/blog.sunix.org.git/_site && jekyll serve --host 0.0.0.0"

Is user able to use --dockerimage and --selector at the same time? If yes, how a user is able to specify component for command?

sunix:
> 
> > `chectl --editor=theia-next`
> 
> I think that next here is a version but not a name of editor, and it is better to use some separator like `/` or `:` to separate name and version. In plugin registry `/` is used for such purposes.
> 
well here, like the language option, it is trying to be simple to use. I was even about to name it `next`. I don't want to confuse user, this is not the registry id

> > chectl devfile:generate  --dockerimage=jekyll/builder --command="mkdir -p /projects/blog.sunix.org.git/_site && jekyll serve --host 0.0.0.0"
> 
> Is user able to use --dockerimage and --selector at the same time? If yes, how a user is able to specify component for command?

yes, the component chosen by the command is arbitrary the first one it founds. Then if it is not the right one, user would simply edit the devfile.

sunix:
> This PR becomes bigger and bigger. I would proceed with merging some part of arguments for `devfile:generate` command and then implement/discuss other arguments in other PRs/issues.
> I think it makes sense to merge at least that part that we demonstrated on DockerCon

what is the problem with merging it in one piece? if it works for everyone ... why loosing time splitting ? I would say 'fail fast' merging it and if there is any issue, we quickly fix it ...

sleshchenko:
> what is the problem with merging it in one piece? if it works for everyone 

No any issues if it works for everyone

README.md
```diff
@@ -76,6 +77,35 @@ EXAMPLES
...
+generate and print a devfile to stdout given some Kubernetes resources and other Che workspaces features (project, language-support, commands etc...)
+
+```
+USAGE
+  $ chectl devfile:generate
+
+OPTIONS
+  -h, --help                 show CLI help
+  --command=command          Command to include in the workspace
+  --dockerImage=dockerImage  dockerImage component to include to the workspace
```
sleshchenko: Maybe it would be better to use `--dockerimage` since it's a value of type for dockerimage component?

README.md
```diff
@@ -76,6 +77,35 @@ EXAMPLES
...
+  --editor=editor            Editor to choose. Currently supported editors: theia-next,theia-1.0.0
```
sleshchenko: It makes sense to specify default value here.

README.md
```diff
@@ -76,6 +77,35 @@ EXAMPLES
...
+  --git-repo=git-repo        Source code git repository to include in the workspace
+
+  --language=language        Add support for a particular language. Currently supported languages:
+                             java,typescript,go,python,c#
+
+  --name=name                Workspace name
```
sleshchenko: Currently, it's not true. It's a devfile name that will be reused for workspace name, and workspace name can be suffixed if workspace with such name already exists.

README.md
```diff
@@ -76,6 +77,35 @@ EXAMPLES
...
+
+  --namespace=namespace      Kubernetes namespace where the resources are defined
+
+  --plugin=plugin            Che plugin to include in the workspace
```
sleshchenko: What is a format of this parameter?

README.md
sunix: ok it is the 'workspace desired name' so ... but the way it is dealing with already existing workspace name is a Che implementation choice: you could choose to fail if already exist. In the devfile, here, user would like to provide the workspace name even if in reality it would be suffixed.

README.md
sunix: at the moment it is json :(

README.md
l0rd: Actually the default is no editor (in the `Devfile`) that is interpreted by Che as "use the default editor component" (currently `theia:next`). Hence I am not sure if and what should be mentioned here as default.

l0rd:
I have fixed CI, addressed some @sleshchenko comments and opened an issue for some stuff that still need to be fixed (#100). @sleshchenko @sunix if you are ok I would merge this PR in its current state.

sunix:
@l0rd thanks very much, was about to have a look to that CI ;) OK to merge !

README.md
```diff
@@ -43,6 +43,7 @@ USAGE
 # Commands
 <!-- commands -->
 * [`chectl autocomplete [SHELL]`](#chectl-autocomplete-shell)
+* [`chectl devfile:generate`](#chectl-devfilegenerate)
```
sleshchenko: It's a minor comment but it would be better to leave `help` command at the beginning.

README.md
```diff
@@ -76,6 +77,37 @@ EXAMPLES
...
+
+  --name=name                Workspace name
+
+  --namespace=namespace      Kubernetes namespace where the resources are defined
+
+  --plugin=plugin            Che plugin to include in the workspace. The format is JSON. For example this is a valid Che
+                             Plugin specification: {"type": "TheEndpointName.ChePlugin", "alias": "java-ls", "id":
+                             "redhat/java/0.38.0"}
+
+  --selector=selector        label selector to filter the Kubernetes resources
```
sleshchenko: Maybe it makes sense to add example to make it clear that selector has format `key=value`:
`--selector=app.kubernetes.io/name=my-app`

README.md
l0rd: @sleshchenko the README is generated automatically by the oclif framework. The flags are ordered alphabetically

README.md
l0rd: Done thanks for the review @sleshchenko

PR #85, opened 2019-04-18
Inject configuration files in all containers if it's not specified

### LINKED ISSUE #59: Target container should be an optional parameter for workspace:inject command
For better UX, if the target container isn't specified for `chectl workspace:inject -k` command, local `kubeconfig` should be injected in every container.

azatsarynnyy:

closes #59 

Executing `chectl workspace:inject` command without a container specified (`-c` option) injects a configuration file in all containers of a Workspace pod.
Tasks are running in parallel indicating the status of each.

See the example of output when injecting kubeconfig in the Workspace with 4 containers.

inject

### DIFF
--- a/src/api/che.ts
+++ b/src/api/che.ts
@@ -70,6 +70,31 @@ export class CheHelper {
     }
   }
 
+  async getWorkspacePodContainers(namespace: string, cheWorkspaceId?: string): Promise<string[]> {
+    this.kc.loadFromDefault()
+    const k8sApi = this.kc.makeApiClient(Core_v1Api)
+
+    const res = await k8sApi.listNamespacedPod(namespace)
+    const pods = res.body.items
+    const wsPods = pods.filter(pod => pod.metadata.labels['che.workspace_id'])
+    if (wsPods.length === 0) {
+      throw new Error('No workspace pod is found')
+    }
+
+    if (cheWorkspaceId) {
+      const wsPod = wsPods.find(p => p.metadata.labels['che.workspace_id'] === cheWorkspaceId)
+      if (wsPod) {
+        return wsPod.spec.containers.map(c => c.name)
+      }
+      throw new Error('Pod is not found for the given workspace ID')
+    } else {
+      if (wsPods.length === 1) {
+        return wsPods[0].spec.containers.map(c => c.name)
+      }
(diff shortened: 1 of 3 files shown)


benoitf:
@azatsarynnyy do you know why on screencast we see `ListrError : Something went wrong` ?

azatsarynnyy:
@benoitf  Listr reports it when some of the subtasks are rejected

src/api/che.ts
```diff
@@ -70,6 +70,31 @@ export class CheHelper {
...
+      const wsPod = wsPods.find(p => p.metadata.labels['che.workspace_id'] === cheWorkspaceId)
+      if (wsPod) {
+        return wsPod.spec.containers.map(c => c.name)
+      }
+      throw new Error('Pod is not found for the given workspace ID')
+    } else {
+      if (wsPods.length === 1) {
+        return wsPods[0].spec.containers.map(c => c.name)
+      }
+      throw new Error('More than one pod with running workspace is found. Please, specify Che Workspace ID.')
```
l0rd: You are asking the user to provide a Workspace ID but it should be always the case here (i.e. `cheWorkspaceId` is not undefined if you are here).

azatsarynnyy: @l0rd workspace id is an optional parameter. If it's omitted, chectl tries to detect it

azatsarynnyy:
@benoitf @l0rd I've updated the PR with one more commit to handling the errors more gracefully.
There're two common situations:
- a file cannot be injected since the `tar` binary isn't present in a target container
- `kubeconfig` already exists in the target container

Now, instead of throwing a `ListrError`, such tasks are just skipped with an explanation of a reason.

image

l0rd:
@azatsarynnyy instead of always show `skipped` for those 2 containers can't we filter them out based on the labels (i.e. not trying to inject at all)?

azatsarynnyy:
> for those 2 containers

@l0rd which of them?
 I see only `che-machine-exec` container is useless here. Agree, we can filter it.

azatsarynnyy:
I think `theia-ide` sidecar also can be a target for the command, e.g. for `gitconfig`
Added filtering of the `che-machine-exec`:
image

PR #86, opened 2019-04-24
Use alias in devfile.yaml and remove component names.

metlos:
Che 7.0.0-beta-4.0 is going to drop support for component names and replaces it with an alias. We need to update the devfile.yaml to take that into account.


l0rd:
@metlos feel free to merge

PR #89, opened 2019-04-28
Adapt chectl to new plugin notation

garagatyi:
Do not merge before https://github.com/eclipse/che/pull/13204 gets merged

### DIFF
--- a/test/api/requests/devfile.invalid
+++ b/test/api/requests/devfile.invalid
@@ -6,10 +6,10 @@ projects:
       type: git
       location: 'git@github.com:che-incubator/chectl.git'
 tools:
-  - name: theia-ide
+  - alias: theia-ide
     type: cheEditor
     id: eclipse/che-theia/1.0.0
-  - name: exec-plugin
+  - alias: exec-plugin
     type: chePlugin
     id: eclipse/che-machine-exec-plugin/0.0.1
 commands:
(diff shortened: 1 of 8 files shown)


test/api/requests/devfile.invalid
```diff
@@ -8,10 +8,10 @@ projects:
 tools:
   - name: theia-ide
```
garagatyi: @metlos isn't this name shouldn't be here anymore?

l0rd: Good catch @garagatyi

garagatyi:
I've merged https://github.com/eclipse/che/pull/13204 but since chectl depends on Che wsmaster nightly image we can merge it tomorrow.
@l0rd can you leave you review, so I would be able to proceed with it?

test/api/requests/devfile.invalid
l0rd: @garagatyi since you have spotted this can you please remove it?

garagatyi: sure

garagatyi: done

garagatyi:
@l0rd just checked it locally - workspace creation/start works as expected with the current nightly.
I'm ready to merge the PR as soon as you approve it

PR #93, opened 2019-05-10
Specify that userland-proxy option should be set to false

apupier:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Binary downloads of `chectl` can be found on [the Release page](https://github.c
 
 Download the `chectl` binary and add it to your PATH.
 
-Currently `chectl` requires [minikube](https://github.com/kubernetes/minikube#installation) and [helm](https://github.com/helm/helm#install) to be locally installed.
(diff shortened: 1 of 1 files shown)


apupier:
otherwise you end up with this message:

```
    × Verify userland-proxy is disabled
      →             Note: you may have to recreate the minikube installation.
      Verify if minikube ingress addon is enabled
      Enable minikube ingress addon
      Retrieving minikube IP and domain for ingress URLs
 »   Error: E_PLATFORM_NOT_COMPLIANT_USERLAND: userland-proxy=false parameter is required on docker daemon but it was
 »   not found.
 »                This setting is given when originally starting minikube. (you can then later check by performing
 »   command : minikube ssh -- ps auxwwww | grep dockerd
 »                It needs to contain --userland-proxy=false
 »                Command that needs to be added on top of your start command:
 »                $ minikube start <all your existing-options> --docker-opt userland-proxy=false
 »                Note: you may have to recreate the minikube installation.
 »
```

l0rd:
@apupier I have added you to the Che organization. You should be able to merge this PR yourself if you like.

PR #96, opened 2019-05-12
updating devfile

sunix:


devfile.yaml
```diff
@@ -1,26 +1,44 @@
 specVersion: 0.0.1
 name: chectl
 projects:
-  - name: chectl
+  -
```
l0rd: That's not a blocker but I am wondering why are you adding a newline here?

devfile.yaml
```diff
@@ -1,26 +1,44 @@
...
 components:
-  - alias: theia-ide
+  - 
+    id: eclipse/che-theia/next
     type: cheEditor
-    id: eclipse/che-theia/1.0.0
-  - type: chePlugin
+  - 
     id: eclipse/che-machine-exec-plugin/0.0.1
+    type: chePlugin
```
l0rd: I don't think that this one is needed. I mean if you omit it it should be automatically added. So let's remove this line and make the devfile 3 lines shorter

devfile.yaml
```diff
@@ -1,26 +1,44 @@
...
+  - 
+    alias: git-devtools
```
l0rd: can we avoid the `alias` here since it's not used in the commands

devfile.yaml
```diff
@@ -1,26 +1,44 @@
...
+    type: dockerimage
+    image: sunix/git-devtools
```
l0rd: Can we use an "official" image here? The main purpose is to avoid maintaining it by ourselves. And I would use an [official `node` image](https://hub.docker.com/_/node/) because we probably want to run `yarn` and `git` from the same terminal.

devfile.yaml
```diff
@@ -1,26 +1,44 @@
...
+    id: che-incubator/typescript/1.30.2
+    type: chePlugin
+    memoryLimit: 1Gi 
+
 commands:
   - name: build
     actions:
       - type: exec
-        component: theia-ide
+        component: ts
```
l0rd: It would be ok to use `ts` component if you hadn't included a `devtools` component as well. Since you have added a `devtool` component, and since a user will mostly use that component to open a terminal, shouldn't we target the `devtool` component to run commands?

devfile.yaml
sunix: for lisibility,
here it is not a big deal, but when you have many elements, it becomes harder to read. This is personal taste :)

devfile.yaml
sunix: ok

devfile.yaml
sunix: not used by commands but it is display nicely with this name on the `my workspace` view

devfile.yaml
sunix: I thought we wanted to avoid 'all in one' images. That one could be reuse in any project and not only nodejs project. I couldn't find any nice official git image. That one is quite cool and include almost all the git commands and extensions + the `tig` command.

devfile.yaml
sunix: it was not a devtools but a git-devtools == git related development tools

devfile.yaml
l0rd: Maybe you can just insert a new line between components (c.f. https://github.com/flathub/flathub/wiki/YAML-Style-Guide)? Your formatting style doesn't look standard and someone with a different "taste" may want to change it as well.

devfile.yaml
l0rd: Ok makes sense

devfile.yaml
l0rd: I don't know what you mean with "all in one" image but we want to avoid running multiple services in one single container. That's why we are separating services in distinct containers. But having one container per CLI tool is bad as well: a user should not switch his terminal to execute `git commit` after executing `yarn test`. The valid argument you have is about `tig` that is not included in the official `node` image. But is the value of `tig` so hight that it justifies maintaining a custom stack? I don't think so...

sunix: Ideally we only need to use che commands to build and test `chectl`. Then you could even test your command in any container. the provided container for git include `tig` but also all the git extension including for instance the ability to run `git add -p` or to clone ssh repo, but these are just 2 examples. The official `node` image does not include git. At least, we don't have a proper support of git in Theia, I am using this daily for the time being.

benoitf:
hello @sunix what is the status of this PR ?

devfile.yaml
l0rd: @sunix it doesn't make sense to have separate containers for every development tool. Mainly because users will need to open a new terminal for every tool they use.

sunix:
Hello i have just updated it (and use it), could you review ? i have added a dev container using che-theia-dev for running yarn and so on.
Let me know if it works for you @benoitf @l0rd

sunix:
@benoitf @l0rd could you review it again ? i have added a dev container (based on che-theia-dev) to perform yarn and yarn test.
Works for you ?

devfile.yaml
benoitf: would be nice here as @l0rd said to not have custom user images

sunix:
looks like i still cannot merge in this repo :(

sunix:
To me it is not a big deal. Running yarn would be done through task che/commands. this is the development mode. When it comes to prepare the commit and push, you use the terminal,  it is another step.
I don't understand why it is a big deal here but not a big deal in our default stacks. All our stacks should not use the default image for building ... for instance, when it is to use maven, we use the container maven:3.6.0-jdk-11 that doesn't contain git ...
But if it really matter, let's create that image but, I would prefere to iterate and do it in another PR as an improvement.

l0rd:
I am ok to merge that and work on a second PR to have a chectl dedicated Dockerfile and image.

> To me it is not a big deal. Running yarn would be done through task che/commands. this is the development mode. When it comes to prepare the commit and push, you use the terminal, it is another step.

If we use one single container instead of 2 we can still do that. Instead with your approach we cannot use a single terminal for both `yarn` and `git`.

> I don't understand why it is a big deal here but not a big deal in our default stacks. All our stacks should not use the default image for building ... for instance, when it is to use maven, we use the container maven:3.6.0-jdk-11 that doesn't contain git ...

That's an issue. I agree with you. But having a dedicated component for each tool is not the solution: it would be a nightmare for people that use the terminal. Better solutions would be to mount dev tools as we want to do with theia plugins runtime or to use images that have the tools (git) as well as the runtime (jdk).

> We can also add ALL the git command line tooling inside the che-theia-dev image.

We can do that of course but `che-theia-dev` image is shared with other projects that may have different requirements (e.g. different version of yarn, node etc...). If someone updates `che-theia-dev` he may break chectl without being aware of it. Having one dedicated chectl image is better.

PR #99, opened 2019-05-13
Support space in paths #95

apupier:
very strangely, only one escaped double-quotes needs to be added.
Cannot add simple double-quotes because the exca library is already
encapsulating inside double quotes.

Signed-off-by: Aurélien Pupier <apupier@redhat.com>

I tried only on Windows.
Worth a check on Linux, I feel like it is working by exploiting a bug in execa libraries more than a clean solution.

### DIFF
--- a/src/installers/helm.ts
+++ b/src/installers/helm.ts
@@ -145,7 +145,7 @@ error: E_COMMAND_FAILED`)
 
   async updateCheHelmChartDependencies(cacheDir: string, execTimeout= 120000) {
     const destDir = path.join(cacheDir, '/templates/kubernetes/helm/che/')
-    await execa.shell(`helm dependencies update --skip-refresh ${destDir}`, { timeout: execTimeout })
+    await execa('helm', ['dependencies', 'update', '--skip-refresh', destDir], { timeout: execTimeout, reject: false})
   }
 
   async upgradeCheHelmChart(flags: any, cacheDir: string, execTimeout= 120000) {
@@ -162,8 +162,18 @@ error: E_COMMAND_FAILED`)
       tlsFlag = `-f ${destDir}values/tls.yaml`
     }
 
-    let command = `helm upgrade --install che --force --namespace ${flags.chenamespace} --set global.ingressDomain=${flags.domain} --set cheImage=${flags.cheimage} --set global.cheWorkspacesNamespace=${flags.chenamespace} ${multiUserFlag} ${tlsFlag} ${destDir}`
-    await execa.shell(command, { timeout: execTimeout })
+    await execa('helm',
+      ['upgrade',
+        '--install', 'che',
+        '--force',
+        '--namespace', flags.chenamespace,
+        '--set', `global.ingressDomain=${flags.domain}`,
+        '--set', `cheImage=${flags.cheimage}`,
+        '--set', `global.cheWorkspacesNamespace=${flags.chenamespace}`,
+        multiUserFlag,
+        tlsFlag,
+        destDir],
+      { timeout: execTimeout, reject: false})
   }
 
   async purgeHelmChart(name: string, execTimeout= 30000) {


l0rd:
@apupier yes this need a proper check and I suspect that we will need to figure out if we are running on windows at runtime and use a special windows command. Looking at the official documentation it looks like you need to use double quotes for all the parameters: https://ss64.com/nt/syntax-esc.html

src/installers/helm.ts
```diff
@@ -145,7 +145,7 @@ error: E_COMMAND_FAILED`)
 
   async updateCheHelmChartDependencies(cacheDir: string, execTimeout= 120000) {
     const destDir = path.join(cacheDir, '/templates/kubernetes/helm/che/')
-    await execa.shell(`helm dependencies update --skip-refresh ${destDir}`, { timeout: execTimeout })
+    await execa.shell(`helm dependencies update --skip-refresh \"${destDir}`, { timeout: execTimeout })
```
l0rd:
```suggestion
    await execa.shell(`"helm" "dependencies" "update" "--skip-refresh" "${destDir}"`, { timeout: execTimeout })
```

src/installers/helm.ts
```diff
@@ -162,7 +162,7 @@ error: E_COMMAND_FAILED`)
       tlsFlag = `-f ${destDir}values/tls.yaml`
     }
 
-    let command = `helm upgrade --install che --force --namespace ${flags.chenamespace} --set global.ingressDomain=${flags.domain} --set cheImage=${flags.cheimage} --set global.cheWorkspacesNamespace=${flags.chenamespace} ${multiUserFlag} ${tlsFlag} ${destDir}`
+    let command = `helm upgrade --install che --force --namespace ${flags.chenamespace} --set global.ingressDomain=${flags.domain} --set cheImage=${flags.cheimage} --set global.cheWorkspacesNamespace=${flags.chenamespace} ${multiUserFlag} ${tlsFlag} \"${destDir}`
```
l0rd:
```suggestion
    let command = `"helm" "upgrade" "--install" "che" "--force" "--namespace" "${flags.chenamespace}" "--set" "global.ingressDomain=${flags.domain}" "--set cheImage=${flags.cheimage}" "--set global.cheWorkspacesNamespace=${flags.chenamespace}" "${multiUserFlag}" "${tlsFlag}" "${destDir}"`
```

src/installers/helm.ts
apupier: this is not working:

```
    × Updating Helm Chart dependencies
      → Error: could not find C:\Users\Aurelien Pupier\AppData\Local\chectl\templates\kubernetes\helm\che": GetFileAttributesEx C:\Users\Aurelien Pupier\Ap
p…
      Deploying Che Helm Chart
Error: Command failed: C:/WINDOWS/system32/cmd.exe /q /s /c ""helm" "dependencies" "update" "--skip-refresh" "C:/Users/Aurelien Pupier/AppData/Local/chectl/templates/kubernetes/helm/che/""
Error: could not find C:/Users/Aurelien Pupier/AppData/Local/chectl/templates/kubernetes/helm/che": GetFileAttributesEx C:/Users/Aurelien Pupier/AppData/Local/chectl/templates/kubernetes/helm/che": The filename, directory name, or volume label syntax is incorrect.
    at makeError (C:/git/chectl/node_modules/execa/index.js:174:9)
    at Promise.all.then.arr (C:/git/chectl/node_modules/execa/index.js:278:16)
```

apupier: was there any specifc reasons to use execa;shell instead of execa directly?
It is working using `await execa('helm', ['dependencies', 'update', '--skip-refresh', destDir], { timeout: execTimeout, reject: false})` (which seems a nice solution)

src/installers/helm.ts
```diff
@@ -162,8 +162,18 @@ error: E_COMMAND_FAILED`)
...
-    await execa.shell(command, { timeout: execTimeout })
+    await execa('helm',
+				['upgrade',
+				'--install', 'che',
+				'--force',
+				'--namespace', flags.chenamespace,
+				'--set', 'global.ingressDomain=${flags.domain}',
+				'--set', 'cheImage=${flags.cheimage}',
+				'--set', 'global.cheWorkspacesNamespace=${flags.chenamespace}',
+				'${multiUserFlag}',
```
apupier: multiUserFlag and tlsFlag are in fact 2 arguments.
i think multiUserFlag is activated by default, if true, it seems that it is working even written this way, withotu the ned to split the arguments.

src/installers/helm.ts
apupier: not working, replaced by using execa instead of execa.shell

benoitf:
hello, is it still an issue ?
`execa` has been updated to v2 as well in the meantime

apupier:
I don't have anymore an environment ready to test it currently.
Closing for now as i don't have plan to investigate in short term.

PR #104, opened 2019-05-21
Fix TLS templates typos

akervern:
Fix few typos for making `global.tls.enabled` work.


l0rd:
@akervern would you be so kind to backport the fix [upstream as well](https://github.com/eclipse/che/blob/master/deploy/kubernetes/helm/che/templates/certificate.yaml#L10)?

PR #112, opened 2019-05-27
Fixes #101: check for minishift platform for minishift addon

### LINKED ISSUE #101: Using the minishift-addon installer with minikube should fail
It should fail before the preflight check and provide a meaningful error. Instead it currently fails trying to install the addon:

```
chectl server:start -a minishift-addon
  √ ✈️  Minikube preflight checklist
    √ Verify if kubectl is installed
    √ Verify if minikube is installed
    √ Verify if minikube is running
    ↓ Start minikube [skipped]
      → Minikube is already running.
    √ Verify userland-proxy is disabled...done.
    √ Verify if minikube ingress addon is enabled
    ↓ Enable minikube ingress addon [skipped]
      → Ingress addon is already enabled.
    √ Retrieving minikube IP and domain for ingress URLs...192.168.1.27.nip.io.
  √ 🏃‍  Running the Che minishift-addon
    √ Apply Che addon...done.
  > ✅  Post installation checklist
    > Che pod bootstrap
      × scheduling
        → ERR_TIMEOUT
        downloading images
        starting
      Retrieving Che Server URL
      Che status check
Error: ERR_TIMEOUT
    at KubeHelper.<anonymous> (C:/snapshot/chectl/lib/api/kube.js:0:0)
    at Generator.next (<anonymous>)
    at fulfilled (C:/snapshot/chectl/node_modules/tslib/tslib.js:104:62)
```

benoitf:
Change-Id: I956a292cca4fbc6a898e83f9afbafd7f90750986

### DIFF
--- a/src/commands/server/start.ts
+++ b/src/commands/server/start.ts
@@ -125,6 +125,13 @@ export default class Start extends Command {
     const operator = new OperatorHelper()
     const minishiftAddon = new MinishiftAddonHelper()
 
+    // matrix checks
+    if (flags.platform === 'minikube' && flags.installer && flags.installer === 'minishift-addon') {
+      this.error(`🛑 Current platform is ${flags.platform }. Minishift addon is only available on top of Minishift platform.`)
+    } else if (flags.platform === 'minishift' && flags.installer && flags.installer === 'helm') {
+      this.error(`🛑 Current platform is ${flags.platform }. Helm installer is only available on top of Minikube platform.`)
+    }
+
     // Platform Checks
     let platformCheckTasks = new Listr(undefined, {renderer: flags['listr-renderer'] as any, collapse: false})
     if (flags.platform === 'minikube') {


src/commands/server/start.ts
```diff
@@ -142,6 +142,13 @@ export default class Start extends Command {
       this.exit()
     }
 
+    if (flags.platform === 'minikube' && flags.installer && flags.installer.includes('minishift')) {
+      platformCheckTasks.add({
```
l0rd: If you want to add a task to the minikube `startTasks` you should do that in `minikube.ts` rather then here.

src/commands/server/start.ts
```diff
@@ -142,6 +142,13 @@ export default class Start extends Command {
...
+        title: '🛑  Minishift Requirement',
+        task: () => this.error(`Current platform is ${flags.platform }. Minishift installers are only available on top of Minishift platform.`)
```
l0rd: Why are you talking about "minishift installers" instead of "minishift-addon installer"? I mean I am not sure how users will interpret "minishift installers": can the operator considered a minishift installer?

src/commands/server/start.ts
benoitf: I didn't want to add it in minikube `startTasks`

AFAIK Minikube startTasks are only for minikube. Here it's more to add a check independent of the runner. Only start() command has the global view of platform and installer

I think there is like a design failure BTW. We should register installer per platform instead of allowing installer globally

I could move it to minikube.ts but IMHO I kept the logic of the start task making the redirect to different implementation based on the flags

l0rd: I see and it makes sense. Then you can just fail without the need of adding a `platformCheckTask`.

src/commands/server/start.ts
benoitf: because I checked for 'minishift' string being included in the installer in case other installers would include minishift in their name

benoitf: but I can be dumb and check for `minishift-addon` directly

src/commands/server/start.ts
benoitf: ok

src/commands/server/start.ts
l0rd: Well we support 3 installers today and one of those is called `minishift-addon`. I don't know if we will ever have other `*minishift*` installers.

Another thing you could add in this PR is a check that `helm` installer is NOT used with the minishift platform. What do you think?

benoitf:
@l0rd PR updated

PR #115, opened 2019-05-27
Remove k8s templates from chectl

### LINKED ISSUE #106: Workspace with mysql fails to start because of outdated helm chart
mysql container requires to be run with non-root user.
Screenshot from 2019-05-24 10-32-33
The following Devfile may be used to reproduce an issue https://gist.githubusercontent.com/sleshchenko/86aa2e010375bc2a472aa3fa6adaaf18/raw/92cb073a2f4124d7445c454e9207a5abc2512f25/petclinic.k8s.devfile.yaml

Helm chart in upstream Che is updated and non-root user is used by default.
Workspace with mysql fails to start because of an outdated helm chart state.

It's pretty easy to copy/paste helm chart from Che Repo but I decided to create this issue to raise a topic of copying/pasting helm chart, only a few people are aware of existence of copy for Che Helm chart in chectl repo and nobody does the same changes which are done for upstream Che.

This issue is about investigation whether it is possible to avoid having a duplicate of Che Helm Chart for Chectl.

benoitf:
Use k8s templates from upstream Eclipse Che

Copy them as post-install step
It fixes #106


l0rd:
I am not sure I understand how postinstall works: the first time `chectl` is started eclipse/che master branch is cloned right? If that's the case I am not sure I want that (it makes the behavior unpredictable, server:start slower and offline-mode even harder than it is today).

What I would like for this kind of issue (sync between the source code of 2 repos) is to automate the creation of an issue each time the upstream code (in this case folder deploy/kubernetes of eclipse/che) is changed.

benoitf:
@l0rd in my PR, postinstall is working as 'yarn install' post operation. So it means that when you clone chectl and run yarn, it will grab all dependencies (including eclipse-che templates)
Then, it's as before, templates will be stored as assets inside the binary with pkg.

l0rd:
@benoitf that's cool. Let me think about it before merging. I need a few hours to get used to the idea :-)

l0rd:
@benoitf to make build reproducible can we use a tag or a commit hash instead of the head of `master` branch?

benoitf:
@l0rd yes we can use commit sha1 as well

benoitf:
@l0rd ok, moved to latest sha1 of master branch of che

PR #119, opened 2019-05-28
Ensure minishift version is 1.33 or higher when using minishift-addon

### LINKED ISSUE #117: Minishift should be v1.33.0 or higher when using the addon
Before [minishift v1.33.0](https://github.com/minishift/minishift/releases/tag/v1.33.0) the minishift addon didn't create the necessary resources to start a Che 7 workspace. 

Hence, when the installer is the minishift-addon, we should check the minishift version and fail if the version is not v1.33.0 or higher.

benoitf:
It Fixes #117

### DIFF
--- a/src/installers/minishift-addon.ts
+++ b/src/installers/minishift-addon.ts
@@ -32,12 +32,34 @@ export class MinishiftAddonHelper {
     }
   }
 
+  static async grabVersion(): Promise<number> {
+    let args = ['version']
+    const { stdout} = await execa('minishift',
+                                     args,
+                                     {reject: false })
+    if (stdout) {
+      return parseInt(stdout.replace(/\D/g, '').substring(0, 3), 10)
+    }
+    return -1
+
+  }
+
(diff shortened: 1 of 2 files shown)


src/installers/minishift-addon.ts
```diff
@@ -61,6 +71,18 @@ export class MinishiftAddonHelper {
     ], {renderer: flags['listr-renderer'] as any})
   }
 
+  async grabVersion(): Promise<number> {
+    let args = ['version']
+    const { stdout} = await execa('minishift',
+                                     args,
+                                     {reject: false })
+    if (stdout) {
+      return parseInt(stdout.replace(/\D/g, '').substring(0, 3), 10)
```
l0rd: Can you please add a test for `grabVersion`. That would really help if we need to refactor or update the parsing mechanism.

benoitf: I've added tests

l0rd: @benoitf thanks

PR #146, opened 2019-06-07
Add MicroK8s as a platform option

thokra:
This adds support for MicroK8s as a possible installation platform. It's mostly just a copy of the minikube setup, with a few extra MicroK8s details.

It will enable the ingress addon if it is disabled, but it will just error out if the storage addon is disabled or if MicroK8s isn't running. 
This is mostly because sudo is required, and I'm not familiar enough with JS/TS to get piping of stdout/stdin to work.

Tested it successfully with microk8s v1.14.2 and helm


benoitf:
somehow I missed that contribution, great

@thokra: AFAIK microk8s is relying on `microk8s.kubectl` command instead of `kubectl` ?

l0rd:
I have tested this PR and could run microk8s. But to do so I have been blocked a few times:

- 1st try failed because kubectl was not found but microk8s.kubectl was installed
image
I fixed that with `sudo snap alias microk8s.kubectl kubectl`
- 2nd try failed at `start microk8s` step. It looks like it didn't found out that microk8s was running (indeed it was) 
image
running chectl a third time fixed this problem
- 3rd try failed at `Enable microk8s addon storage` step. The addon was not installed initially, chectl installed it and checked again if it was enabled immediately after. But this addon takes a few seconds to be enabled and chectl fails.
image
waiting 30s and running chectl again fixed the problem
- 4th try failed at `Verify if Helm is installed step`. That's not related to this PR at all but I wanted to track it.
image
running `sudo snap install helm --classic` fixed the problem
- 5th try was successful. Hurray!!

Am I +1 to merge this PR as it is (because it adds value and doesn't do any harm) and fix the issues with other PRs.

@thokra would you be able to resolve the conflicts so that we can proceed and merge?

l0rd:
Created https://github.com/che-incubator/chectl/issues/196 to improve microk8s support

PR #162, opened 2019-06-12
devfile:generate use the latest version of language extensions

sunix:


benoitf:
yesterday I had internal 500 errors on github when trying to open it :-/

Issue #14, opened 2018-12-11
Support Kubernetes embedded in Docker for mac


skabashnyuk:
Minikube running on Virtualbox is quite a heavy. It would be nice to allow using built-in Kubernetes in Docker for Mac.

l0rd:
You can/should use the [hyperkit driver](https://github.com/kubernetes/minikube/blob/master/docs/drivers.md#hyperkit-driver) if you are on macOS

l0rd:
Anyway yes I have asked @benoitf if he could work on that

PR #170, opened 2019-06-14
Add initial support for openshift, k8s and docker-desktop platforms

benoitf:
`--platform=openshift` , `--platform=k8s` or `--platform=docker-desktop`
It fixes https://github.com/che-incubator/chectl/issues/14 and fixes https://github.com/che-incubator/chectl/issues/80

note: you need to be logged in on the underlying platform before running chectl

I've tested by using Openshift v3.11 on RHPDS and on K8S/digitalocean and Docker Desktop for mac

example with command
```$ chectl server:start --platform=docker-desktop```
image

note: using operator is still failing on workspaces due to https://github.com/che-incubator/chectl/issues/92

Change-Id: Ic541c5d7932c6782c67bc0103a0d941fd34288d6

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ EXAMPLES
   $ chectl autocomplete --refresh-cache
 ```
 
-_See code: [@oclif/plugin-autocomplete](https://github.com/oclif/plugin-autocomplete/blob/v0.1.0/src/commands/autocomplete/index.ts)_
+_See code: [@oclif/plugin-autocomplete](https://github.com/oclif/plugin-autocomplete/blob/v0.1.1/src/commands/autocomplete/index.ts)_
 
 ## `chectl devfile:generate`
 
@@ -124,7 +124,7 @@ OPTIONS
   --all  see all commands in CLI
 ```
 
-_See code: [@oclif/plugin-help](https://github.com/oclif/plugin-help/blob/v2.1.6/src/commands/help.ts)_
+_See code: [@oclif/plugin-help](https://github.com/oclif/plugin-help/blob/v2.2.0/src/commands/help.ts)_
 
 ## `chectl server:delete`
 
@@ -167,7 +167,7 @@ OPTIONS
   -o, --cheboottimeout=cheboottimeout      (required) [default: 40000] Che server bootstrap timeout (in milliseconds)
 
   -p, --platform=platform                  [default: minikube] Type of Kubernetes platform. Valid values are "minikube",
-                                           "minishift".
+                                           "minishift", "k8s", "openshift".
 
   -s, --tls                                Enable TLS encryption and multi-user mode
 
--- a/src/commands/server/start.ts
+++ b/src/commands/server/start.ts
@@ -21,8 +21,11 @@ import { KubeHelper } from '../../api/kube'
 import { HelmHelper } from '../../installers/helm'
 import { MinishiftAddonHelper } from '../../installers/minishift-addon'
 import { OperatorHelper } from '../../installers/operator'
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: src/commands/server/start.ts [hunk 1]
```diff
@@ -141,6 +152,18 @@ export default class Start extends Command {
       this.error(`🛑 Current platform is ${flags.platform }. Minishift addon is only available on top of Minishift platform.`)
     } else if (flags.platform === 'minishift' && flags.installer && flags.installer === 'helm') {
       this.error(`🛑 Current platform is ${flags.platform }. Helm installer is only available on top of Minikube platform.`)
+    } else if (flags.platform === 'openshift') {
+      if (flags.installer) {
+        if (flags.installer === 'helm') {
+          this.error(`🛑 Current platform is ${flags.platform }. Helm installer is only available on top of Minikube platform.`)
```
l0rd: "Helm installers is only available on top of Minikube platform should be changed" in "...on top of Minikube and Kubernetes platforms"

src/platforms/k8s.ts
```diff
@@ -0,0 +1,70 @@
...
+      {
+        title: 'Verify if kubectl is installed',
+        task: () => {
+          if (!commandExists.sync('kubectl')) {
+            command.error('E_REQUISITE_NOT_FOUND')
+          }
+        }
+      },
+      { title: 'Verify remote kubernetes status',
+        task: async (_ctx: any, task: any) => {
```
l0rd: Could we use something like https://github.com/che-incubator/chectl/blob/master/src/commands/server/stop.ts#L59 to avoid some adherence to `kubectl`?

src/platforms/openshift.ts
```diff
@@ -0,0 +1,57 @@
...
+          }
+        }
+      },
+      // Should automatically compute route if missing
+      { title: 'Verify domain is set',
+        task: (_ctx: any, task: any) => {
+          if (flags.domain === undefined || flags.domain === '') {
+            command.error('E_MISSING_ARGUMENT: the domain parameter needs to be defined.')
+          }
+          task.title = `${task.title}...set to ${flags.domain}.`
```
l0rd: I saw in your screenshot that the domain is set to ".". How did it worked?

src/commands/server/start.ts
benoitf: fixed

benoitf: I've changed the logic to be based on the installer instead of the platform

src/platforms/k8s.ts
benoitf: yes, done

benoitf: BTW I still check for kubectl tool as it's used sometimes in start sequences depending on the installer

src/platforms/openshift.ts
benoitf: screenshot is updated

benoitf:
PR updated to be reviewed again with addition of `docker desktop` support as well now

README.md
```diff
@@ -167,7 +167,7 @@ OPTIONS
   -o, --cheboottimeout=cheboottimeout      (required) [default: 40000] Che server bootstrap timeout (in milliseconds)
 
   -p, --platform=platform                  [default: minikube] Type of Kubernetes platform. Valid values are "minikube",
-                                           "minishift".
+                                           "minishift", "k8s", "openshift".
```
l0rd: you need to regenerate it to add `docker-desktop`

l0rd:
Tested on minikube and docker-desktop and works fine. Nice job @benoitf

Issue #114, opened 2019-05-27
Implement and ability to start che server with custom devfile registry


skabashnyuk:
Implement and ability to start che server with custom devfile registry
https://github.com/eclipse/che-devfile-registry

sleshchenko:
@metlos Can it be closed after your PRs?

PR #189, opened 2019-06-24
feat: Add support for overriding plugin registry url and devfile registry url at the server start.

metlos:
As the most controversial thing in this PR I've changed the dependency on eclipse-che (from which the helm charts are extracted at build time) to the master branch instead of a fixed (and old) commit.

While this makes the build not reproducible, it also makes us always use the latest helm chart instead of its outdated snapshot.

This implements #114


README.md
```diff
@@ -191,12 +191,18 @@ OPTIONS
 
   -t, --templates=templates                [default: templates] Path to the templates folder
 
+  --devfile-registry=devfile-registry      [default: https://che-devfile-registry.openshift.io/] The URL of the Devfile
```
sleshchenko: We're going to implement an ability to deploy Devfile/Plugin registry along with Che with an argument, like `--deploy-devfile-registry`. it would set needed URL automatically
Also, we're going to implement auto propagation of env vars that starts with `CHE_` to Che Server, it would allow to set any value to plugin/devfile registry env var.
^ It's what deploy_che.sh script does


@metlos Do you think these arguments can be still useful? Won't it be confusing to have two `--devfile-registry` and `--deploy-devfile-registry` (this can be named differently parameters?

metlos: Well, given the fact that people are already playing with custom registries (e.g. https://github.com/eclipse/che/issues/13634), I think it is going to be useful to have the ability to preconfigure the server with a custom registry url.

l0rd: I think it makes sense to have both parameters `--devfile-registry-url` and `--deploy-devfile-registry`.

l0rd: To make it more explicit I would name the parameter `deploy-devfile-registry-url`

package.json
```diff
@@ -25,7 +25,7 @@
     "axios": "^0.19.0",
     "cli-ux": "^5.2.1",
     "command-exists": "^1.2.8",
-    "eclipse-che": "git://github.com/eclipse/che#9976bb0",
+    "eclipse-che": "git://github.com/eclipse/che#master",
```
l0rd: I think that it makes sense to use `che#master` here on `chectl#master` branch since we plan to release chectl as we do with che (at every sprint) and we should then use `che#$release-num` on `chectl#release-num` branch.

l0rd: cc @benoitf

README.md
metlos: Why `deploy-`? I agree `devfile-registry-url` would be more explicit and would explain the purpose more cleanly, but the `deploy-` prefix muddies the meaning for me again :)

I like @sleshchenko's idea for having `--deploy-devfile-registry=(true|false)` to deploy the devfile registry locally but IMHO that is out of scope of this issue.

But given the above, `--deploy-devfile-registry-url` is just not clear to me - am I deploying something using it or am I setting some configuration property? or both?

package.json
benoitf: I'm also fine for master. I was using it but @l0rd said to use a fixed sha-1

To provide more content here: what I believe is that chectl should have helm charts for the nightlies and also embed the helm charts for every previous Che releases. (like the previous docker-based CLI)
So it can work smoothly on every Che server (it will provide the associate charts for the matching version)
And a chectl needs to be released every night as well.

README.md
metlos: I've renamed the arguments to `--devfile-registry-url` and `--plugin-registry-url`. 

I'm looking forward to having `--deploy-*-registry` but I took the liberty of not doing that in the scope of this PR.

l0rd:
@metlos what should we do with this PR? I am +1 to merge if you can rebase it

metlos:
I've successfully rebased this and also updated the operator to support configuring the devfile registry url (which has already been merged). Let me retest this once more before merging.

l0rd:
@metlos if you squash your commits take into account that we are using https://www.conventionalcommits.org/en/v1.0.0-beta.2/ (c.f. the Semantic Pull Request check)

PR #202, opened 2019-07-03
fix: update che API devfile endpoint

l0rd:
Fixes https://github.com/eclipse/che/issues/13623

### DIFF
--- a/test/api/che.test.ts
+++ b/test/api/che.test.ts
@@ -17,6 +17,7 @@ const namespace = 'che'
 const workspace = 'workspace-0123'
 const cheURL = 'https://che-che.192.168.64.34.nip.io'
 const devfileServerURL = 'https://devfile-server'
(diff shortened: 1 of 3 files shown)


test/api/che.test.ts
```diff
@@ -130,7 +130,7 @@ describe('Che helper', () => {
         .get('/devfile.yaml')
         .replyWithFile(200, __dirname + '/requests/devfile.valid', { 'Content-Type': 'text/plain; charset=utf-8' }))
       .nock(cheURL, api => api
-        .post('/api/devfile')
+        .post('/api/workspace/devfile')
```
benoitf: Maybe use a constant for the API_ENDPOINT

l0rd: :+1: done

benoitf: thx

PR #211, opened 2019-07-10
fix(k8s): token missing in endpoint when calling /healthz

benoitf:
It fixes https://github.com/eclipse/che/issues/13800

Grab secret from serviceAccount and use it when calling the endpoint

I've tested on a remote k8s and docker-desktop

Change-Id: I5daf5df41d8cebc51ef501c629eca65bc63ae729

### DIFF
--- a/src/api/kube.ts
+++ b/src/api/kube.ts
@@ -791,18 +791,74 @@ export class KubeHelper {
     return this.kc.getCurrentContext()
   }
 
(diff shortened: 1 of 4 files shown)


benoitf:
@l0rd ok, I moved it to a separate method, it is better.

l0rd:
Nice!

PR #216, opened 2019-07-15
fix(workspace): fix workspace:start command with authentication enabled

benoitf:
Adds a flag to provide the token
It fixes https://github.com/eclipse/che/issues/13828

```
$ chectl workspace:start --access-token=${JWT_TOKEN} -f <(./bin/run devfile:generate --name=minimal-wksp)
  ✔ Retrieving Che Server URL...http://che-che.192.168.99.127.nip.io
  ✔ Verify if Che server is running...RUNNING (auth enabled)
  ✔ Create workspace from Devfile /dev/fd/63

Workspace IDE URL:
http://che-che.192.168.99.127.nip.io/dashboard/#/ide/florent/minimal-wksp
```

Change-Id: I05e27dab4409624e29648a9660106e4711676064

### DIFF
--- a/src/commands/workspace/start.ts
+++ b/src/commands/workspace/start.ts
@@ -42,10 +42,20 @@ export default class Start extends Command {
       description: 'workspace name: overrides the workspace name to use instead of the one defined in the devfile. Works only for devfile',
       required: false,
     }),
+    'access-token': string({
+      description: 'Che OIDC Access Token',
+      env: 'CHE_ACCESS_TOKEN'
+    }),
     'listr-renderer': string({
       description: 'Listr renderer. Can be \'default\', \'silent\' or \'verbose\'',
       default: 'default'
     }),
+  }
+
+  async checkToken(flags: any, ctx: any) {
+    if (ctx.isAuthEnabled && !flags['access-token']) {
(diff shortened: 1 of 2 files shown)


src/commands/workspace/start.ts
```diff
@@ -42,12 +42,22 @@ export default class Start extends Command {
...
+    }),
     'listr-renderer': string({
       description: 'Listr renderer. Can be \'default\', \'silent\' or \'verbose\'',
       default: 'default'
     }),
   }
 
+  async checkToken(flags: any, ctx: any) {
+    if (ctx.isAuthEnabled && !flags['access-token']) {
+      this.error('E_AUTH_REQUIRED - Che authentication is enabled and an access token need to be provided (flag --access-token).')
```
sunix: Maybe tell the user in few words, how to get that token ?

benoitf: there is a link in server:stop command but it's about che6 doc
https://www.eclipse.org/che/docs/che-6/authentication.html

I would say that we should have like openshift a 'copy login' command or something like that from dashboard UI

sunix: Could you create an issue for that (adding a 'copy login' from dashboard UI) ?

benoitf: https://github.com/eclipse/che/issues/13852

Issue #174, opened 2019-06-17
Which platform should be default ?


benoitf:
Now that there are several platforms options on chectl, I would be in favor of making this parameter explicitly defined.

Because it's not because I've `minikube` or `minishift` tools installed locally that I want to install che locally.

So I'm thinking it would be nice that if we don't specify, we have the available platforms and for local platforms a quick info if the tool is there

let say, if `minishift` is found, then it's displayed as a possible value or with a special colors or two differents lists. If `minikube` is not found then I see that it's not fully possible, etc.

PR #217, opened 2019-07-15
Always ask platform parameter. And not assume minikube is the default one.

benoitf:
You may have minikube installed but maybe it's not where you want to use the cli tool.
Here it makes --platform parameter mandatory

#174


benoitf:
I think it's important for 7.0.0 and it's not a so deep change :-)

rhopp:
To me (personally) this doesn't seem like very important issue (it will be critical for future productization - but this can happen after GA). I'm +0 for merging this one.

PR #221, opened 2019-07-16
fix(yarn.lock): Old templates were used as yarn.lock is locking sha-1 master

benoitf:
Self-update `eclipse/che` and `eclipse/che-operator` dependencies when running yarn install command.
So yarn.lock is using latest master versions every time it's used.

Change-Id: I430aa868b07cbc954de699154ad070b3cca42756

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,8 @@
   "scripts": {
     "postinstall-helm": "rimraf templates/kubernetes && cpx 'node_modules/eclipse-che/deploy/kubernetes/**' 'templates/kubernetes'",
     "postinstall-operator": "rimraf templates/che-operator && cpx 'node_modules/eclipse-che-operator/deploy/**' 'templates/che-operator'",
-    "postinstall": "npm run -s postinstall-helm & npm run -s postinstall-operator",
+    "postinstall-repositories": "yarn upgrade eclipse-che eclipse-che-operator",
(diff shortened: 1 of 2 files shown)


l0rd:
@benoitf so this means that even setting `master` branch on https://github.com/che-incubator/chectl/blob/master/package.json#L29 is always overridden from yarn.lock?

benoitf:
@l0rd yes correct this is why I didn't noticed some changes as well on my environment

l0rd:
Anyway that's a pity that we don't have a way to just "enforce" master

benoitf:
we could on CI enfore the update of some dependencies (so it will update yarn.lock automatically)

benoitf:
@davidfestal you fear something but now even without my PR your change will never be used so I'm not sure about the required changes that you're expecting

Except then I could do yarn commands but only on CI

davidfestal:
@benoitf I was meaning that you would have to redo a similar PR very soon when my PR is merged. But if you want you still can merge this one first.

benoitf:
ok so now I've added a command that force the update of the cloned dependencies

benoitf:
@davidfestal I also have updated to the latest sha1 of che-operator but it should be done automatically as well now

l0rd:
> ok so now I've added a command that force the update of the cloned dependencies

Nice. That's a good idea.

PR #222, opened 2019-07-16
Default generated (by devfile:generate) dockerimage should not override the container image entrypoint

sunix:
See https://github.com/eclipse/che/issues/13817#issuecomment-511747894

A lot of the images would have a entrypoint define that has to be executed. For instance some may have permission settings for arbitrary user in the entrypoint. To avoid breaking few containers that would run on a Che workspace, this is just generating the dockerimage with the right CMD to avoid kubernetes to run the app but keeping the original entrypoint call.

### DIFF
--- a/src/commands/devfile/generate.ts
+++ b/src/commands/devfile/generate.ts
@@ -129,8 +129,7 @@ export default class Generate extends Command {
         image: `${flags.dockerimage}`,
         memoryLimit: '512M',
         mountSources: true,
-        command: ['tail'],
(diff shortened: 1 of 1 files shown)


sunix:
ok you are right let's keep like it is.

benoitf:
@sunix maybe the generated devfile can be commented and explains the goal of the line (it's to override entry point)

sunix:
@benoitf yeah i was thinking the same, that for every generated pieces we could add a comment with a link to the documentation of that specific part

PR #224, opened 2019-07-17
fix(helm): Automatically grab tls email from che-tls and checks for cert-manager

benoitf:
For now tls let's encrypt certificates are failing as the default email is with example.com
But as we require che-tls secret with email inside, re-use this one (if `--tls` option is given)

And when cert-manager is not installed and that `--tls` option is enabled there are helm errors hard to understand for newcomers, so detect it automatically if cert-manager has been installed or not and provides guidance.

It completes fixes of https://github.com/eclipse/che/issues/13801

### DIFF
--- a/src/api/kube.ts
+++ b/src/api/kube.ts
@@ -973,6 +973,24 @@ export class KubeHelper {
     }
   }
 
+  async getSecret(name = '', namespace = 'default'): Promise<string | undefined> {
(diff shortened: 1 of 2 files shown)


benoitf:
it's part of https://github.com/eclipse/che/issues/13801 with 7.0.0

PR #228, opened 2019-07-19
remove devfile and plugin registries defaults

sparkoo:
Registries will be deployed locally by default (https://github.com/eclipse/che/issues/13557), so we don't want to have default registries URL sent to helm.

- [x] Wait for related PR: https://github.com/eclipse/che/pull/13890

This will now work just with helm chart. It must be synced with work on operator side.

### DIFF
--- a/src/installers/helm.ts
+++ b/src/installers/helm.ts
@@ -195,18 +195,30 @@ error: E_COMMAND_FAILED`)
 
     let multiUserFlag = ''
     let tlsFlag = ''
-    let setOptions = ''
+    let setOptions = []
 
     if (flags.multiuser) {
       multiUserFlag = `-f ${destDir}values/multi-user.yaml`
     }
 
     if (flags.tls) {
-      setOptions = `--set global.cheDomain=${flags.domain} --set global.tls.email='${ctx.tlsEmail}'`
+      setOptions.push(`--set global.cheDomain=${flags.domain} --set global.tls.email='${ctx.tlsEmail}'`)
       tlsFlag = `-f ${destDir}values/tls.yaml`
     }
 
-    let command = `helm upgrade --install che --force --namespace ${flags.chenamespace} --set global.ingressDomain=${flags.domain} ${setOptions} --set cheImage=${flags.cheimage} --set global.cheWorkspacesNamespace=${flags.chenamespace} --set che.workspace.devfileRegistryUrl=${flags['devfile-registry-url']} --set che.workspace.pluginRegistryUrl=${flags['plugin-registry-url']} ${multiUserFlag} ${tlsFlag} ${destDir}`
+    if (flags['plugin-registry-url']) {
+      setOptions.push(`--set che.workspace.pluginRegistryUrl=${flags['plugin-registry-url']} --set chePluginRegistry.deploy=false`)
+    }
+
+    if (flags['devfile-registry-url']) {
+      setOptions.push(`--set che.workspace.devfileRegistryUrl=${flags['devfile-registry-url']} --set cheDevfileRegistry.deploy=false`)
+    }
+
+    setOptions.push(`--set global.ingressDomain=${flags.domain}`)
+    setOptions.push(`--set cheImage=${flags.cheimage}`)
+    setOptions.push(`--set global.cheWorkspacesNamespace=${flags.chenamespace}`)
+
+    let command = `helm upgrade --install che --force --namespace ${flags.chenamespace} ${setOptions.join(' ')} ${multiUserFlag} ${tlsFlag} ${destDir}`
 
     let {code, stderr} = await execa.shell(command, { timeout: execTimeout, reject: false })
     // if process failed, check the following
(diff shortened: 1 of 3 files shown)


benoitf:
FYI the semantic check is failing / commit is not matching https://www.conventionalcommits.org/en/v1.0.0-beta.4/

src/installers/helm.ts
```diff
@@ -195,18 +195,26 @@ error: E_COMMAND_FAILED`)
...
     }
 
     if (flags.tls) {
-      setOptions = `--set global.cheDomain=${flags.domain} --set global.tls.email='${ctx.tlsEmail}'`
+      setOptions.push(`--set global.cheDomain=${flags.domain} --set global.tls.email='${ctx.tlsEmail}'`)
       tlsFlag = `-f ${destDir}values/tls.yaml`
     }
 
-    let command = `helm upgrade --install che --force --namespace ${flags.chenamespace} --set global.ingressDomain=${flags.domain} ${setOptions} --set cheImage=${flags.cheimage} --set global.cheWorkspacesNamespace=${flags.chenamespace} --set che.workspace.devfileRegistryUrl=${flags['devfile-registry-url']} --set che.workspace.pluginRegistryUrl=${flags['plugin-registry-url']} ${multiUserFlag} ${tlsFlag} ${destDir}`
+    if (flags['plugin-registry-url']) {
```
benoitf: what is happening if you use server:start without `--plugin-registry-url` ? it is failing no ?

sparkoo: it doesn't set helm parameter so value from helm chart is used. By default registries will be deployed in your minikube and URLs will be set to those https://github.com/eclipse/che/pull/13890

benoitf: ok it was not clear enough for me that this PR can only be merged after https://github.com/eclipse/che/pull/13890

sparkoo: well, I don't think that this must be merged after. Helm charts currently does not set the registry urls, but then `che.properties` values are used, which are again set to `openshift.io` ones https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L580
So basically nothing happens when we merge this one. The chectl registries url params was added just 2 days ago https://github.com/che-incubator/chectl/pull/189

sparkoo:
> FYI the semantic check is failing / commit is not matching https://www.conventionalcommits.org/en/v1.0.0-beta.4/

fixed. Would be good to know this is enabled in README and not to hit it in PR. Chectl issues are in this repository or are they going to be moved to root Che repo?

benoitf:
issues are in che repository.

I will update PR template and README

src/installers/helm.ts
```diff
@@ -195,18 +195,26 @@ error: E_COMMAND_FAILED`)
...
+      setOptions.push(`--set che.workspace.pluginRegistryUrl=${flags['plugin-registry-url']} --set chePluginRegistry.deploy=false`)
+    }
+
+    if (flags['devfile-registry-url']) {
+      setOptions.push(`--set che.workspace.devfileRegistryUrl=${flags['devfile-registry-url']} --set cheDevfileRegistry.deploy=false`)
+    }
+
+    let command = `helm upgrade --install che --force --namespace ${flags.chenamespace} --set global.ingressDomain=${flags.domain} ${setOptions.join(' ')} --set cheImage=${flags.cheimage} --set global.cheWorkspacesNamespace=${flags.chenamespace} ${multiUserFlag} ${tlsFlag} ${destDir}`
```
l0rd: @sparkoo since you are here don't you mind pushing every `--set <param>` to `setOptions` so that the command get more readable?

sparkoo: yep, done

sparkoo:
@benoitf could you please approve if you are now happy with changes?

l0rd:
@sparkoo you should be able to merge whenever you are happy with your PR

l0rd:
I have tested it and it worked fine. The only thing is that I expected that we wait for the registries bootstrap in the `Post installation checklist` as we do for che, postgres and keycloak. We can add that in a separate PR as well but that would make the UX much better.

sparkoo:
> I have tested it and it worked fine. The only thing is that I expected that we wait for the registries bootstrap in the `Post installation checklist` as we do for che, postgres and keycloak. We can add that in a separate PR as well but that would make the UX much better.

done https://github.com/che-incubator/chectl/pull/228/commits/3ac7552435d900ddd091b67b18a04325d198055c

l0rd:
@sparkoo thanks. One last thing: you should also stop the registries (if you find them) in the [`server:stop` command](https://github.com/che-incubator/chectl/blob/master/src/commands/server/stop.ts).

sparkoo:
yes, now, when I've changed the selectors, it does not even work 
```
  ✖ Wait until Che pod is deleted
    → Get pods by selector "app=che" returned 3 pods (1 was expected)
```

I'm on it

sparkoo:
@l0rd `server:stop` fixed

PR #234, opened 2019-07-24
docs(README): when we committed changes we didn't push README.MD changes

benoitf:
### What does this PR do?
Update README with latest automatic changes
Probably need to check if README is modified in PR check to avoid merging PR without README.MD changes

Change-Id: Id53ea7000641e9aea22795b1c8de5d823a831f55


benoitf:
@sparkoo I am as well :-)
this is why I'm thinking about a check that can only let merge if README is up-to-date

PR #237, opened 2019-07-26
Adapt chectl to fixes for TLS for helm chart

sleshchenko:
### What does this PR do?
**TODO** revise `--chenamespace=chenamespace` parameter

1. Fixes documentation since `--tls` does not enable multiuser anymore.
2. The only server strategy that works is `multi-host` and it requires wildcards certificates. It's why helm chart does not include cluster issuer and certificates anymore. User must provide `che-tls` secret with certificate content.
It's why checking that cert-manager is set is removed, as well as propagating `tls.email` parameter.

It's tested on GCP with the following configurations:
```
./bin/run server:start -a=helm -p=k8s -b=che.gcps.my-ide.cloud --tls --multiuser
```
```
./bin/run server:start -a=helm -p=k8s -b=che.gcps.my-ide.cloud --tls
```
```
./bin/run server:start -a=helm -p=k8s -b=che.gcps.my-ide.cloud --multiuser
```

It's needed after https://github.com/eclipse/che/pull/13946

### DIFF
--- a/src/installers/helm.ts
+++ b/src/installers/helm.ts
@@ -37,7 +37,7 @@ export class HelmHelper {
           const kh = new KubeHelper()
           const exists = await kh.secretExist('che-tls', `${flags.chenamespace}`)
           if (!exists) {
-            throw new Error(`TLS option is enabled but che-tls secret does not exist in '${flags.chenamespace}' namespace. Example on how to create the secret with TLS: kubectl create secret tls che-tls --namespace=che --key=privkey.pem --cert=fullchain.pem`)
+            throw new Error(`TLS option is enabled but che-tls secret does not exist in '${flags.chenamespace}' namespace. Example on how to create the secret with TLS: kubectl create secret tls che-tls --namespace='${flags.chenamespace}' --key=privkey.pem --cert=fullchain.pem`)
           }
           task.title = `${task.title}...che-tls secret found.`
         }
--- a/src/installers/helm.ts
+++ b/src/installers/helm.ts
@@ -33,38 +33,13 @@ export class HelmHelper {
         enabled: () => {
           return flags.tls
         },
-        task: async (ctx: any, task: any) => {
-          const kh = new KubeHelper()
-          const exists = await kh.secretExist('che-tls')
-          if (!exists) {
-            throw new Error('TLS option is enabled but che-tls secret does not exist in default namespace. Example on how to create the secret: kubectl create secret generic che-tls --from-literal=ACME_EMAIL=my@email-address.com')
-          }
-          const tlsEmail = await kh.getSecret('che-tls')
-          if (tlsEmail === undefined) {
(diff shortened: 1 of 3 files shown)


l0rd:
I am on PTO I won't able to test it today. Will be back on Monday but feel free to merge it if it gets approved before.

src/installers/helm.ts
```diff
@@ -35,38 +35,13 @@ export class HelmHelper {
         },
         task: async (ctx: any, task: any) => {
           const kh = new KubeHelper()
-          const exists = await kh.secretExist('che-tls')
+          const exists = await kh.secretExist('che-tls', `${flags.chenamespace}`)
           if (!exists) {
-            throw new Error('TLS option is enabled but che-tls secret does not exist in default namespace. Example on how to create the secret: kubectl create secret generic che-tls --from-literal=ACME_EMAIL=my@email-address.com')
+            throw new Error(`TLS option is enabled but che-tls secret does not exist in '${flags.chenamespace}' namespace. Example on how to create the secret: kubectl create secret generic che-tls --from-literal=ACME_EMAIL=my@email-address.com`)
```
benoitf: It seems it is not any longer with ACME_EMAIL ( example provided)

benoitf:
FYI semantic commits and DCO checks are failing

benoitf:
(see CONTRIBUTING.MD)

src/installers/helm.ts
sleshchenko: Good catch. Changed to 
```
throw new Error(`TLS option is enabled but che-tls secret does not exist in '${flags.chenamespace}' namespace. Example on how to create the secret with TLS: kubectl create secret tls che-tls --namespace=che --key=privkey.pem --cert=fullchain.pem`)
```
Do you think it's good enough?

src/installers/helm.ts
```diff
@@ -33,38 +33,13 @@ export class HelmHelper {
...
-          const exists = await kh.apiVersionExist('certmanager.k8s.io')
+          const exists = await kh.secretExist('che-tls', `${flags.chenamespace}`)
           if (!exists) {
-            throw new Error(`TLS option is enabled but cert-manager API has not been found. Cert Manager is probably not installed. Example on how to install it:
-            $ kubectl create namespace cert-manager
-            $ kubectl label namespace cert-manager certmanager.k8s.io/disable-validation=true
-            $ kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v0.8.1/cert-manager.yaml --validate=false
-
-            Please install cert-manager.`)
+            throw new Error(`TLS option is enabled but che-tls secret does not exist in '${flags.chenamespace}' namespace. Example on how to create the secret with TLS: kubectl create secret tls che-tls --namespace=che --key=privkey.pem --cert=fullchain.pem`)
```
l0rd: In the example command to create the secret you can replace `--namespace=che` with `--namespace=${flags.chenamespace}` so that the example uses the right namespace.

sleshchenko: fixed

PR #241, opened 2019-08-02
fix(operator): fix deploying on K8s with operator and TLS 

sleshchenko:
### What does this PR do?
fix(operator): fix deploying on K8s with operator and TLS 

https://github.com/eclipse/che/issues/14102

### DIFF
--- a/src/api/kube.ts
+++ b/src/api/kube.ts
@@ -760,6 +760,9 @@ export class KubeHelper {
       yamlCr.spec.server.cheImageTag = imageAndTag.length === 2 ? imageAndTag[1] : 'latest'
       yamlCr.spec.auth.openShiftoAuth = flags['os-oauth']
       yamlCr.spec.server.tlsSupport = flags.tls
+      if (flags.tls) {
+        yamlCr.spec.k8s.tlsSecretName = 'che-tls'
+      }
       yamlCr.spec.server.selfSignedCert = flags['self-signed-cert']
       yamlCr.spec.k8s.ingressDomain = flags.domain
       let pluginRegistryUrl = flags['plugin-registry-url']
--- a/README.md
+++ b/README.md
@@ -210,8 +210,10 @@ OPTIONS
   -p, --platform=platform                      [default: minikube] Type of Kubernetes platform. Valid values are
                                                "minikube", "minishift", "k8s", "openshift", "microk8s".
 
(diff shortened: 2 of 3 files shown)


src/platforms/minishift.ts
```diff
@@ -48,14 +48,6 @@ export class MinishiftHelper {
       },
       // { title: 'Verify minishift memory configuration', skip: () => 'Not implemented yet', task: () => {}},
       // { title: 'Verify kubernetes version', skip: () => 'Not implemented yet', task: () => {}},
-      { title: 'Retrieving minishift IP and domain for routes URLs',
```
benoitf: what if you use helm or minishift addon on minishift ?

sleshchenko: AFAIK minishift addon does not need domain as well.

About helm - I'm not even sure that it's a working case at all.
If we need it - we should adapt helm chart to be run onto OpenShift.
Like, use Routes instead of Ingresses, set OpenShift as Che Server infra because K8s infra uses namespaces instead of project and it's supposed to fail.

sleshchenko: BTW Chectl does not allow it 
Screenshot_20190802_123403

src/api/kube.ts
```diff
@@ -759,6 +759,9 @@ export class KubeHelper {
       yamlCr.spec.server.cheImageTag = imageAndTag.length === 2 ? imageAndTag[1] : 'latest'
       yamlCr.spec.auth.openShiftoAuth = flags['os-oauth']
       yamlCr.spec.server.tlsSupport = flags.tls
+      if (flags.tls) {
+        yamlCr.spec.k8s.tlsSecretName = `che-tls`
```
rhopp: Where this `che-tls` value comes from?
What happens, if I put my own value in CheCluster YAML provided by parameter `--che-operator-cr-yaml`. To me it seems, that it will be overriden.

davidfestal: @rhopp 

Then nothing will be overriden and additional chectl parameters will be ignored

That's the reason of line [752](https://github.com/che-incubator/chectl/pull/241/files/eae1e1a3f14078c159b5770d6ad6c3a3c3b9e413#diff-a468c62edd386b77d72847cae8716b18R752)

benoitf:
@sleshchenko FYI it doesn't work on CI (format style error)

sleshchenko:
@benoitf Thanks! Fixed.

PR #244, opened 2019-08-07
fix(openshift/minishift): Do ask for domain parameter for openshift/minishift platforms

sleshchenko:
### What does this PR do?
Do ask for domain parameter for openshift/minishift platforms

This PR is under WIP phase because it is supposed to be merged in 7.1.0

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,8 +179,8 @@ USAGE
 OPTIONS
   -a, --installer=installer                    Installer type. Valid values are "helm", "operator" and "minishift-addon"
 
-  -b, --domain=domain                          Domain of the Kubernetes/OpenShift cluster (e.g.
-                                               starter-us-east-2.openshiftapps.com or <local-ip>.nip.io)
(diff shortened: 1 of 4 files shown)


benoitf:
hello, what is related issue ? to know if it's for `7.0.0` or `7.1.0` ?

sleshchenko:
@benoitf It's a tiny improvement I discovered during using chectl.
I have not even created an issue because it would be bigger than the fix itself.
Can you set milestone directly to PR or I should create an issue?
I'm OK with `7.1` since it is not blocker, but it should not break anything - I believe `7.0` should be fine as well.

l0rd:
@sleshchenko why are you removing setting the domain automatically on minishift?

sleshchenko:
@l0rd Because it is unused on OpenShift platform. 
OS infra assign host automatically for Routes.
For K8s it's needed because there Che Server defines host based rules for Ingresses.

l0rd:
@sleshchenko ok makes sense

PR #251, opened 2019-08-14
WIP Revert "userland-proxy=false is not required anymore"

sleshchenko:
### What does this PR do?
This reverts commit b862e0ededde7efa90f4c1d7f599ffbb1fa4aa96.

It follows the following email posted to `che-dev` https://www.eclipse.org/lists/che-dev/msg03375.html

Which minishift/minikube problems this PR fixes:
1. [NodeJS + Mongo Devfile](https://github.com/eclipse/che-devfile-registry/blob/master/devfiles/nodejs-mongo/devfile.yaml#L54) is working again. It can be fixed in an alternative way, and use `localhost` for application->DB communication.
2. If a user defines in their Devfile multiple Deployments and Service for setting up communication - it will work again. Like an example https://github.com/redhat-developer/devfile/blob/master/samples/web-nodejs-with-db-sample/devfile.yaml
3. [Discoverable endpoints](https://github.com/eclipse/che/blob/caae00e21b2fd316560a406f2e1d8ad98edf1dad/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/devfile/Constants.java#L67
) work again. User may declare them in dockerimage component. As an example, the [NodeJS+Mongo Devfile](https://github.com/eclipse/che-devfile-registry/blob/master/devfiles/nodejs-mongo/devfile.yaml#L54) referenced above.

:warning: It's has WIP in the title because it was created to get feedback and concerns about requiring `userland-proxy=false` again

https://github.com/eclipse/che/issues/14009

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,9 @@ _See code: [src/commands/server/delete.ts](https://github.com/che-incubator/chec
 
 ## `chectl server:start`
 
-start Eclipse Che Server
+start Eclipse Che Server.
+
+Currently `chectl` requires [minikube](https://github.com/kubernetes/minikube#installation) and [minishift](https://github.com/minishift/minishift) launched with _--docker-opt userland-proxy=false_ option.
 
 ```
 USAGE
--- a/src/platforms/minikube.ts
+++ b/src/platforms/minikube.ts
@@ -45,6 +45,22 @@ export class MinikubeHelper {
         },
         task: () => this.startMinikube()
       },
+      { title: 'Verify userland-proxy is disabled',
+        task: async (_ctx: any, task: any) => {
+          const userlandDisabled = await this.isUserLandDisabled()
+          if (!userlandDisabled) {
+            command.error(`E_PLATFORM_NOT_COMPLIANT_USERLAND: userland-proxy=false parameter is required on docker daemon but it was not found.
+            This setting is given when originally starting minikube. (you can then later check by performing command : minikube ssh -- ps auxwwww | grep dockerd
+            It needs to contain --userland-proxy=false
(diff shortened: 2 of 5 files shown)


benoitf:
I think you have valid arguments but other previous valid arguments were that restarting minishift or minikube with these settings need mostly to recreate the VMs and from UX it's not so nice to not be able to use the instance already there

sleshchenko:
@benoitf 

> I think you have valid arguments but other previous valid arguments were that restarting minishift or minikube with these settings need mostly to recreate the VMs and from UX it's not so nice to not be able to use the instance already there

Should we print add like the following warning?
```
If you do specify `userland-proxy=false` then discoverable endpoints[see doc] in workspaces won't work. You can skip this checking by `--skip-userland-proxy-check`
```
Or maybe we even are able to do interactive option:
```
If you do specify `userland-proxy=false` then discoverable endpoints[see doc] in workspaces won't work.
Type `Y` if you are want to proceed without discoverable endpoints functionality:
```

Messages are just examples and definitely must be better.

benoitf:
@sleshchenko  is that che infra is able to check that a user is adding discoverable endpoints in its devfile, etc and then report that if `userland-proxy=false` it won't work as well ?

sleshchenko:
> @sleshchenko is that che infra is able to check that a user is adding discoverable endpoints in its devfile, etc and then report that if userland-proxy=false it won't work as well ?

Good idea, but I doubt that it can check it. chectl could provide this info for Che Server with env var.
BTW user then may face this issue with their own K8s components with multiple deployments and services

PR #255, opened 2019-08-19
Various small fixes

sleshchenko:
### What does this PR do?
It contains some minor fixes that are separate by self-describing commits.
Here I'm adding some context to some of them:
1. fix: fix formatting in ts files
It's a pretty weird thing when you modify a few lines in a file, use `Format document` feature and some extra lines are modified as well with formatting fixes only.
I performing formatting of code, and hopefully, if everyone won't merge unformatted code, then we won't get such situation anymore.
I tried to automate a process of formatting but failed because of the following issues 
https://github.com/microsoft/TypeScript/issues/11629
https://github.com/vvakame/typescript-formatter/issues/118

2. fix: Fixes checking if Che Server is available
`isCheServerReady` method requires che URL and at the same time checks if namespace already exists. 
And CheURL is supposed to be fetched with a dedicated method that requests resources from k8s/OS API, and it does not check che namespace exists.
I believe that namespace existence is moved in the right place in the fix.

I've found these issues during work on https://github.com/che-incubator/chectl/issues/77
And created a separate PR to make review easier.

### DIFF
--- a/src/commands/workspace/start.ts
+++ b/src/commands/workspace/start.ts
@@ -77,8 +77,8 @@ export default class Start extends Command {
       {
         title: 'Verify if Che server is running',
         task: async (ctx: any, task: any) => {
-          if (!await che.isCheServerReady(ctx.cheURL, flags.chenamespace)) {
-            this.error(`E_SRV_NOT_RUNNING - Che Server is not running.\nChe Server cannot be found in Kubernetes Namespace "${flags.chenamespace}". Have you already start it?\nFix with: start Che server: chectl server:start\nhttps://github.com/eclipse/che`, { code: 'E_SRV_NOT_RUNNNG' })
+          if (!await che.isCheServerReady(ctx.cheURL)) {
+            this.error(`E_SRV_NOT_RUNNING - Che Server is not available by ${ctx.cheURL}`, { code: 'E_SRV_NOT_RUNNNG' })
           }
           const status = await che.getCheServerStatus(ctx.cheURL)
           ctx.isAuthEnabled = await che.isAuthenticationEnabled(ctx.cheURL)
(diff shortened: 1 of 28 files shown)


src/commands/workspace/start.ts
```diff
@@ -77,8 +77,8 @@ export default class Start extends Command {
       {
         title: 'Verify if Che server is running',
         task: async (ctx: any, task: any) => {
-          if (!await che.isCheServerReady(ctx.cheURL, flags.chenamespace)) {
-            this.error(`E_SRV_NOT_RUNNING - Che Server is not running.\nChe Server cannot be found in Kubernetes Namespace "${flags.chenamespace}". Have you already start it?\nFix with: start Che server: chectl server:start\nhttps://github.com/eclipse/che`, { code: 'E_SRV_NOT_RUNNNG' })
+          if (!await che.isCheServerReady(ctx.cheURL)) {
+            this.error(`E_SRV_NOT_RUNNING - Che Server is not running.\nChe Server is not available by ${ctx.cheURL}. T_O_`, { code: 'E_SRV_NOT_RUNNNG' })
```
l0rd: what's `T_O_`?

sleshchenko: Thanks for catching it.
It's TODO hidden from TODO bot that is actually disabled in this repo )
So, will revise this message and remove TODO

sleshchenko: TODO is removed and the message is changed to 
`E_SRV_NOT_RUNNING - Che Server is not available by ${ctx.cheURL}`

sleshchenko:
I'm merging this PR since other changes depend on that but Dear Reviewers feel free to comment after the merge and I will create another PR with fixes.

PR #258, opened 2019-08-22
fix(kube): fix catching of errors

sleshchenko:
### What does this PR do?
fix catching of errors

It does not really solve https://github.com/eclipse/che/issues/14303 but helps to get a cause of failing

### DIFF
--- a/src/api/kube.ts
+++ b/src/api/kube.ts
@@ -51,7 +51,8 @@ export class KubeHelper {
         })
       }
     } catch (e) {
-      throw new Error(e.body.message)
+      if (e.body && e.body.message) throw new Error(e.body.message)
+      else throw new Error(e)
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/api/kube.ts
```diff
@@ -189,7 +195,8 @@ export class KubeHelper {
       const options = new V1DeleteOptions()
       await k8sCoreApi.deleteNamespacedRole(name, namespace, options)
     } catch (e) {
-      throw new Error(e.body.message)
+      if (e.body && e.body.message) throw new Error(e.body.message)
```
benoitf: should we use helper method ? Instead of duplicating the same lines ?

sleshchenko: :+1:

sleshchenko:
@benoitf Please take a look again

PR #257, opened 2019-08-21
Refactor chectl to make tasks reusable for different commands

sleshchenko:
### What does this PR do?
This PR has two major purposed:
- Refactor chectl to make tasks reusable for different commands;
- Implement an ability to start already deployed but stopped Che.

The main idea of refactoring is the following:
Have the following different packages:
- `api` <- holds everything that might be reused while working with different APIs BUT not tasks, cli stuff here
- `tasks` <- holds tasks that are devided into different classes by components 
  | - installers\
  | - platforms\
  | che
- `commands` <- holds commands of chectl, command should reuse tasks from `tasks` package or it can declare it's own unique small tasks

The main pros of this refactoring: encapsulate logic related to some component to one class,
like currently, all commands read `flags['che-namespace']` everytime when they need to do something with che components (find ingress, scale deployment) but it could be encapsulated in one `che-tasks` class.

Note that this PR also contains changes how chectl check if there is already deployed Che, after that - with them it's possible to start already deployed but stopped Che.

This PR also contains a lot of small fixes but I did no manage to highlight them and move to separate commits, like:
- `platform` parameter for `server:start` command is required only when there is no already Deployed Che;
- `server:stop` now stops devfile and plugin registry as well as Che, Postgres and Keycloak;
- print a warning that multiuser is automatically deployed when `che-operator` is used;
- minishift addon is updated to use Deployment instead of DeploymentConfig, so chectl get rid of calling DC methods;

### Is this PR tested?
I made sure that E2E tests are not broken but there are not much covered cases yet.

I tried to cover start-stop-start scenario with E2E tests https://github.com/sleshchenko/chectl/compare/startStopStartScenario...sleshchenko:startStopStartScenarioE2E?expand=1
but they failed because of the current implementation of `api/kube.ts#waitUntilPodIsDeleted` method, it says that it waits until the pod is deleted but in fact, it waits until pod is not ready. It causes an issue that `server:start` gets two Che pods intead of expected 1. I'll register a separate issue for it.

I did manual testing and here I'm posting detailed results. It may be useful to review them as well since it shows what is the actual output of chectl:

#### No Platform Specified
<details>
<summary>Screenshot</summary>

Screenshot_20190910_135245

</details>

##### Minishift Addon

  <details>
<summary>server:start --platform=minishift</summary>

Screenshot_20190910_141404

</details>

  <details>
<summary>server:stop</summary>

Screenshot_20190910_145520

</details>

  <details>
<summary>server:start</summary>

[...]

PR #301, opened 2019-09-06
Small fixes

sleshchenko:
### What does this PR do?
This PR contains some extracted changes from https://github.com/che-incubator/chectl/pull/257
Commits are self-explaining, so please review it in commit-by-commit manner.

https://github.com/che-incubator/chectl/pull/257

### DIFF
--- a/src/installers/helm.ts
+++ b/src/installers/helm.ts
@@ -12,7 +12,7 @@ import { Command } from '@oclif/command'
 import * as commandExists from 'command-exists'
 import * as execa from 'execa'
 import * as fs from 'fs'
-import { mkdirp } from 'fs-extra'
+import { mkdirp, remove } from 'fs-extra'
 import * as Listr from 'listr'
 import { ncp } from 'ncp'
 import * as path from 'path'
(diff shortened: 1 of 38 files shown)


src/installers/helm.ts
```diff
@@ -182,6 +182,7 @@ error: E_COMMAND_FAILED`)
   async prepareCheHelmChart(flags: any, cacheDir: string) {
     const srcDir = path.join(flags.templates, '/kubernetes/helm/che/')
     const destDir = path.join(cacheDir, '/templates/kubernetes/helm/che/')
+    await remove(destDir)
```
l0rd: That makes sense but shouldn't we do the same for every installer (operator and minishift addon)?

sleshchenko: Let me check. I faced this issue only with helm installer.

sleshchenko: You're totally right!
I thought that the error happened in `mkdirp` but actually an error happens when there are some conflicting files.
Screenshot_20190823_145921

I'm adding `removeDir` for operator and minishift addon as well.

src/flags.ts
```diff
@@ -0,0 +1,34 @@
+/*********************************************************************
```
l0rd: Maybe this file should be better named: `common-flags.ts`?

src/installers/helm.ts
sleshchenko: Fixed in https://github.com/che-incubator/chectl/pull/301/commits/0b01a4166c490d3a1f6e8be23ea761f80ab1e9ec

src/flags.ts
sleshchenko: Fixed.

PR #312, opened 2019-09-11
Fix missing crc flag check

gorkem:
Adds crc platform flag check as an option to
namespace creation.

Fixes https://github.com/eclipse/che/issues/14516


gorkem:
@benoitf it should be good to go now

benoitf:
seems semantic check is failing

commit message should be:

```
fix(crc): missing flag check 
```

PR #335, opened 2019-10-02
fix(inject): Rely on che.original_name label when looking for a workspace pod

azatsarynnyy:

### What does this PR do?
Fixes looking for a workspace pod while `workspace:inject`.

closes https://github.com/eclipse/che/issues/14746

### DIFF
--- a/src/api/che.ts
+++ b/src/api/che.ts
@@ -39,7 +39,7 @@ export class CheHelper {
 
     const res = await k8sApi.listNamespacedPod(namespace)
     const pods = res.body.items
-    const wsPods = pods.filter(pod => pod.metadata!.labels!['che.workspace_id'])
+    const wsPods = pods.filter(pod => pod.metadata!.labels!['che.original_name'] === 'che-workspace-pod')
     if (wsPods.length === 0) {
       throw new Error('No workspace pod is found')
     }
@@ -64,7 +64,7 @@ export class CheHelper {
 
     const res = await k8sApi.listNamespacedPod(namespace)
     const pods = res.body.items
-    const wsPods = pods.filter(pod => pod.metadata!.labels!['che.workspace_id'])
+    const wsPods = pods.filter(pod => pod.metadata!.labels!['che.original_name'] === 'che-workspace-pod')
     if (wsPods.length === 0) {
       throw new Error('No workspace pod is found')
     }
--- a/src/api/che.ts
+++ b/src/api/che.ts
@@ -39,7 +39,7 @@ export class CheHelper {
 
     const res = await k8sApi.listNamespacedPod(namespace)
     const pods = res.body.items
-    const wsPods = pods.filter(pod => pod.metadata!.labels!['che.original_name'] === 'che-workspace-pod')
+    const wsPods = pods.filter(pod => pod.metadata!.labels!['che.workspace_id'] && pod.metadata!.labels!['che.original_name'] !== 'che-jwtproxy')
     if (wsPods.length === 0) {
       throw new Error('No workspace pod is found')
     }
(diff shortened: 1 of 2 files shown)


src/api/che.ts
```diff
@@ -39,7 +39,7 @@ export class CheHelper {
 
     const res = await k8sApi.listNamespacedPod(namespace)
     const pods = res.body.items
-    const wsPods = pods.filter(pod => pod.metadata!.labels!['che.workspace_id'])
+    const wsPods = pods.filter(pod => pod.metadata!.labels!['che.original_name'] === 'che-workspace-pod')
```
benoitf: we have these labels with all installs ? operator,helm and minishift ?

azatsarynnyy: @benoitf good question )
@sleshchenko do you happen to know?

benoitf: hum in fact it's ok, as it's che server that is creating these workspaces so we don't care about install part

benoitf:
@azatsarynnyy you need to fix commit message to be compliant with semantic commits

like:
```
fix(workspace:inject): Rely on che.original_name label when looking for a workspace pod 
```

and fix the build

src/api/che.ts
sleshchenko: I doubt that we can rely on `che-workspace-pod` value...
As you can see - it's used only in PluginBroker component related class https://github.com/eclipse/che/search?q=che-workspace-pod&unscoped_q=che-workspace-pod
And only if environment do not have already some pod https://github.com/eclipse/che/blob/dd47af60fdc569b4770659514c0c88b0f26184a3/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesPluginsToolingApplier.java#L106

I assume it may not work if you use devfile with k8s/OS/dockerimage component

azatsarynnyy:
After discussing it with @sleshchenko offline, we've figured out that the problem is actually related to detecting a HOME dir. In some containers (e.g. jwtproxy) chectl is failed detecting HOME.

We decided to skip jwtproxy pod and possibly all the containers where chectl cannot detect HOME dir during `workspace:inject` command execution. I'll update the PR later.

It may be improved once https://github.com/eclipse/che/issues/14752 is fixed.

benoitf:
thanks for the  updated report

apupier:
I think the error message needs to be updated too:
from `More than one pod with running workspace is found. Please, specify Che Workspace ID.` to `More than one pod with running workspace is found. Please, specify Che Workspace name.`
https://github.com/che-incubator/chectl/blob/db55f947c7230d12acd9d830c416040063ee76b0/src/api/che.ts#L57
https://github.com/che-incubator/chectl/blob/db55f947c7230d12acd9d830c416040063ee76b0/src/api/che.ts#L82
https://github.com/che-incubator/chectl/blob/master/test/api/che.test.ts#L239

sleshchenko:
@apupier Workspace name won't help since one workspace has multiple pods: main and jwtproxy.

apupier:
> @apupier Workspace name won't help since one workspace has multiple pods: main and jwtproxy.

I'm confused then. What needs to be provided? The pod names? (so not the Workspace ID, not the Workspace name)

azatsarynnyy:
> What needs to be provided?

A workspace ID has to be provided.
ATM it works with single-user Che only.

azatsarynnyy:
Once this PR is done, it will be fixed for multi-user as well.

apupier:
> A workspace ID has to be provided.

How to retrieve a workspace ID?
In UI there is a "Workspace name" is it the same thing?

Screenshot from 2019-10-08 16-21-08

azatsarynnyy:
The PR has been updated

PR #353, opened 2019-10-15
Added an ability to update deployed Che with operator installer

sleshchenko:
### What does this PR do?
Adds an ability to update Che deployed with an operator by executing a command `server:update`.
It's not a final state where we want `server:update` be but it's a good start point where we can start using it.

https://github.com/eclipse/che/issues/14793

##### 7.0.0 to nightly
1. Download chectl released version like [7.0.0](https://github.com/che-incubator/chectl/releases/tag/20190814170217), unpack it, remove `~/.local/share/chectl/client`, and make sure that chectl version returns correct version. And only then install Che with this chectl.
2. Remove `~/.local/share/chectl/client` and use chectl from this branch (.tar.gz for your OS can be found in the following [archive](https://dropmefiles.com/osUs7) for chectl:nightly for my branch) and do `/bin/run server:update --installer=operator --platform=minishift`
3. Modify CR and null all images that are there.
4. Make sure that Che is fully updated. 

The latest try https://youtu.be/7hh4V3V3GFU

##### 6.19.6 to 7.2.0
1. Deploy Che 6.19.6 with operator but without chectl (did not find the corresponding chectl version for it):
  1. Clone https://github.com/sleshchenko/che-operator
  2. Checkout to 6.19.6 (it's based on 7.0.0-rc-4.0 you can check the latest commits to understand what I changed to get 6.19.6 deployed https://github.com/sleshchenko/che-operator/commits/6.19.6)
  3. Current dir should be newly clonned repo
  4. Execute the following commands(if facts it's the same that chectl does to deploy che with CheOperator but without platform checks):
```bash
cd deploy

# addon admin-user must be installed! Or use another user with admin rights
oc login -u admin  
oc new-project che
oc apply -f service_account.yaml
oc apply -f role.yaml
oc apply -f cluster_role.yaml
oc apply -f cluster_role_binding.yaml
oc apply -f role_binding.yaml

oc apply -f ./crds/org_v1_che_crd.yaml

oc apply -f ./operator.yaml
oc apply -f ./crds/org_v1_che_cr.yaml
```
  5. Wait until Che is ready to use.
2. You may be interested in running Che 6, Che 7 Preview workspaces. So DO IT - run them.
3. Download chectl 7.2.0 binaries with this PR changes https://dropmefiles.com/ksOMB.
4. Unpack version for your OS
5. Execute `rm -r ~/.local/share/chectl` to avoid caching issues.
6. Execute `./bin/chectl server:update --platform=minishift --installer=operator`
7. Wait until new operator is ready.
7. Wait until new operator finishes his work and restart everything it needs.
8. Edit CR:
  - set empty values for all images, like che-server, keycloack, plugin-registry.
  - set `useExternalDevfileRegistry: false`
9. Wait until new Che is ready.
10. Test existing running workspaces.
11. Try to run new Che 7 workspaces.

My laptop is not powerful enough or my minishift has not enough memory. So minishift crashed during update, minishift stop/start does not help to make it working again https://youtu.be/IBzCrRpr-3Y

On OpenShift 4.2 on AWS test scenario is done successfully, see screencast https://youtu.be/7Jqfxi8NvyU (sorry for not sound there)

##### Versions comparing
Now we print the following message
```
Found deployed Che with operator [eclipse/che-operator]:7.0.0
      You are going to update it to [eclipse/che-operator]:7.2.0
Note that che operator will update components images (che server, plugin registry) only if their values
      are not overridden in eclipse-che Customer Resource. So, you may need to remove them manually.
      Press q to quit or any key to continue
```
And user has to decide by themselves if an update should be performed or not.

- [ ] implement some auto versions comparing, like to not allow 7.2.0 upgrading to 7.0.0 since it supposed by fail because of DB migrations.
- [ ] confirmation needs some additional attention from the user. We could consider adding colored text for confirmation like with https://github.com/klaussinani/signale or https://github.com/chalk/chalk. I did not implement it in this PR since it may postpone feature delivering because of CQs
- [ ] show a information about default images for new version of che-operator

##### Wait until the updating process is done
Currently after updating all additional .yaml resources like SA, Role, RoleBinding we:
  - update **kubectl.kubernetes.io/restartedAt** (it's the same logic as kubectl does for rollout restart)
  - wait 1 second to let k8s lauch new replicaset
  - wait until new deployment does not have any unavailable replica
  - we do not track Che Server, Keycloak pods at all.

- [ ] it would be better to expose particular pod status, like we do on `server:start` (scheduled, pulling, pulled)
- [ ] we should consider tracking that new Che Operator triggers rolling updates for components like (keycloak, che server, plugin registry) and wait until they are finished. Also, when we do `server:update` and there is not CR changes - like we updating nightly version to nightly, then chectl probably needs to trigger Che components deployments to do rollout.

##### Updating to nightly
Che Operator in master branch contains the latest released tag as default one. As a result if you update to nightly - you'll get che-operator nightly but all che components will have the latest released version https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L25.
It should be improved

### DIFF
--- a/src/commands/server/update.ts
+++ b/src/commands/server/update.ts
@@ -9,36 +9,145 @@
  **********************************************************************/
 
 import { Command, flags } from '@oclif/command'
+import { boolean, string } from '@oclif/parser/lib/flags'
+import { cli } from 'cli-ux'
+import * as fs from 'fs-extra'
+import * as Listr from 'listr'
+import * as notifier from 'node-notifier'
+import * as path from 'path'
 
-import { cheNamespace, listrRenderer } from '../../common-flags'
+import { cheDeployment, cheNamespace, listrRenderer } from '../../common-flags'
+import { CheTasks } from '../../tasks/che'
+import { InstallerTasks } from '../../tasks/installers/installer'
+import { K8sTasks } from '../../tasks/platforms/k8s'
+import { PlatformTasks } from '../../tasks/platforms/platform'
 
 export default class Update extends Command {
   static description = 'update Eclipse Che Server'
 
   static flags = {
-    help: flags.help({ char: 'h' }),
+    installer: string({
+      char: 'a',
+      description: 'Installer type',
+      options: ['helm', 'operator', 'minishift-addon'],
+      default: ''
+    }),
+    platform: string({
+      char: 'p',
+      description: 'Type of Kubernetes platform. Valid values are \"minikube\", \"minishift\", \"k8s (for kubernetes)\", \"openshift\", \"crc (for CodeReady Containers)\", \"microk8s\".',
+      options: ['minikube', 'minishift', 'k8s', 'openshift', 'microk8s', 'docker-desktop', 'crc'],
+    }),
     chenamespace: cheNamespace,
-    'listr-renderer': listrRenderer
+    templates: string({
+      char: 't',
+      description: 'Path to the templates folder',
+      default: Update.getTemplatesDir(),
+      env: 'CHE_TEMPLATES_FOLDER'
+    }),
+    'che-operator-image': string({
+      description: 'Container image of the operator. This parameter is used only when the installer is the operator',
+      default: 'quay.io/eclipse/che-operator:nightly'
+    }),
+    'skip-version-check': boolean({
+      description: 'Skip user confirmation on version check',
+      default: false
+    }),
+    'deployment-name': cheDeployment,
+    'listr-renderer': listrRenderer,
+    help: flags.help({ char: 'h' }),
+  }
+
+  static getTemplatesDir(): string {
+    // return local templates folder if present
+    const TEMPLATES = 'templates'
+    const templatesDir = path.resolve(TEMPLATES)
+    const exists = fs.pathExistsSync(templatesDir)
+    if (exists) {
+      return TEMPLATES
+    }
+    // else use the location from modules
+    return path.join(__dirname, '../../../templates')
+  }
+
+  checkIfInstallerSupportUpdating(flags: any) {
+    // matrix checks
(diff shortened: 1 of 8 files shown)


sleshchenko:
cc @l0rd @benoitf @nickboldt @davidfestal I think it's a point which we can deliver as the initial implementation of `server:update` but we definitely should improve it later. So, please review. I've attached screencast of my testing, and also build archive for all OS are available if you want to play with it as well.

src/commands/server/update.ts
```diff
@@ -9,36 +9,141 @@
...
     chenamespace: cheNamespace,
-    'listr-renderer': listrRenderer
+    templates: string({
+      char: 't',
+      description: 'Path to the templates folder',
+      default: Update.getTemplatesDir(),
+      env: 'CHE_TEMPLATES_FOLDER'
+    }),
+    'skip-version-check': boolean({
+      description: 'Should be specified if user confirmation after version check should be skipped',
```
mshaposhnik: maybe `Skip user confirmation on version check` ?

src/commands/server/update.ts
```diff
@@ -9,36 +9,141 @@
...
+      title: '↺  Updating...',
+      task: () => new Listr(installerTasks.updateTasks(flags, this))
+    })
+
+    try {
+      const ctx: any = {}
+      await preInstallTasks.run(ctx)
+
+      if (!ctx.isCheDeployed) {
+        this.error('There is no found deployed Che. Deploy before updating')
```
skabashnyuk: what about
```Eclipse Che deployment is not found. Use chectl server:start to initiate new deployment.```?

PR #368, opened 2019-10-25
[WIP] che-14963 CodeReady specific version of chectl

ibuziuk:
### What does this PR do?
Creating `codeready` specific version of `chect` for CRW 2.0
PR is meant to be merged in the dedicated `crw` branch. What does this PR do:

- [x] Changing labels from `che` to `codeready`
- [ ] Removing support of helm & minishift addon
- [ ] Updating help for `workspace` and `server` commands

https://github.com/eclipse/che/issues/14963

### DIFF
--- a/src/api/che.ts
+++ b/src/api/che.ts
@@ -108,7 +108,7 @@ export class CheHelper {
   }
 
   async cheOpenShiftURL(namespace = ''): Promise<string> {
-    const route_names = ['che', 'che-host']
+    const route_names = ['codeready', 'che-host']
     for (const route_name of route_names) {
       if (await this.oc.routeExist(route_name, namespace)) {
         const protocol = await this.oc.getRouteProtocol(route_name, namespace)
--- a/src/tasks/che.ts
+++ b/src/tasks/che.ts
@@ -28,20 +28,20 @@ export class CheTasks {
   cheNamespace: string
 
   cheAccessToken: string
-  cheSelector = 'app=che,component=che'
+  cheSelector = 'app=codeready,component=codeready'
   cheDeploymentName: string
 
   keycloakDeploymentName = 'keycloak'
-  keycloakSelector = 'app=che,component=keycloak'
+  keycloakSelector = 'app=codeready,component=keycloak'
 
   postgresDeploymentName = 'postgres'
-  postgresSelector = 'app=che,component=postgres'
+  postgresSelector = 'app=codeready,component=postgres'
 
   devfileRegistryDeploymentName = 'devfile-registry'
-  devfileRegistrySelector = 'app=che,component=devfile-registry'
+  devfileRegistrySelector = 'app=codeready,component=devfile-registry'
 
   pluginRegistryDeploymentName = 'plugin-registry'
-  pluginRegistrySelector = 'app=che,component=plugin-registry'
+  pluginRegistrySelector = 'app=codeready,component=plugin-registry'
 
   constructor(flags: any) {
     this.kube = new KubeHelper(flags)


benoitf:
it's like how it has to be maintained ? a crw branch ? it looks like quite `touchy`

I thought you only need a build script like we do for the release
https://github.com/che-incubator/chectl/blob/master/make-release.sh

so you run against current source and you've a codeready compliant binary


Also right now crw is based against master so it won't work well AFAIK

ibuziuk:
@nickboldt @l0rd I have opened this draft PR with the checklist of what needs to be done for CRW specific chectl and make sure that we are on the same page. 

@nickboldt could you clarify what is the role of https://github.com/redhat-developer/codeready-workspaces-deprecated for CRW 2.0:
- is the repo deprecated or should we use it somehow for chectl?
- some yaml files were embedded in the script `deploy.sh` https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/master/operator-installer/deploy.sh#L222 should those be reused by codeready specific chectl
- Also I see https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/master/operator-installer/custom-resource.yaml was added yesterday, should it be also used by chectl ?

nickboldt:
> is the repo deprecated or should we use it somehow for chectl?

It was deprecated as I was under the impression that:

* for 2.0, the stacks would no longer be needed, so the stack deps would no longer need to be built
* for 2.0, we would not need deploy.sh 

The first assumption is no longer true; we're reusing the stacks as devfile runtime containers AND in many cases also reusing them as LS sidecar containers.

The second assumption is also not true, as we need SOME way to support deployment to OCP/OSD 3.11. Short of forcing people to install OLM and then use the operator, we need to have some kind of deploy.sh / crwctl mechanism. Even with a forked crwctl I would still provide a deploy.sh script, but it would simply be a wrapper for the ctl command, so as to make it more obvious which flags to use. This will also be documented in the CRW 2 docs, but it's easier to explain to the Docs team what to write when it's "copy the commands from the scriptlet" :)

So maybe we need a new name for this repo, but I don't want to blindly refactor stuff (which would break a lot of jobs / create a lot of work). At the very least I updated the repo description last week:

"This repo is for CodeReady Workspaces 1.x and 2, in support of non-operator based deployment (OCP 3.11, etc.)"

I'll update the root README too. 

> some yaml files were embedded in the script deploy.sh https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/master/operator-installer/deploy.sh#L222 should those be reused by codeready specific chectl

At this point they're not being used for CRW 2 deployment; ideally we would need none of that for 2.0. Ideally we could run the whole deployment from crwctl + crw.yaml file, with no extra embeds/files/metadata/cruft.

> Also I see https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/master/operator-installer/custom-resource.yaml was added yesterday, should it be also used by chectl ?

It was added a couple weeks ago, and yes, the current approach to using vanilla `chectl` is documented here [0], using this cr.yaml [1]:

[0] https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/master/operator-installer/README.adoc#installing-with-chectl-draft
[1] https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/master/operator-installer/custom-resource.yaml

ibuziuk:
> it's like how it has to be maintained ? a crw branch ? it looks like quite touchy

This is meant to be a temp solution for CRW 2.0

> Also right now crw is based against master so it won't work well AFAIK

`crw` branch is planned to be built on top of 7.3.1

> I thought you only need a build script like we do for the release
https://github.com/che-incubator/chectl/blob/master/make-release.sh
so you run against current source and you've a codeready compliant binary

@benoitf what about  `help` and unsupported `codeready` flavors (helm / minishift addon) ?

ibuziuk:
> while I'd prefer an approach where we can use vanilla chectl + a new flag --selector=codeready to override default che selector, +1 to this quick hack approach. If we're doing a proper fork, we would want to remove kube/helm support too.

@nickboldt after discussing with @l0rd it was decided that just updating selectors is not enough. We do need to update `help` and drop unsupported flavours at very least

nickboldt:
> @nickboldt after discussing with @l0rd it was decided that just updating selectors is not enough. We do need to update help and drop unsupported flavours at very least

+1, agreed.

l0rd:
@benoitf my idea was to temporary host a CRW version of chectl (crwctl) on a branch to unblock QE and as a kind of PoC. Then we should: 
1) create a script that automatically updated upstream chectl sources in to crwctl sources 
2) have the crwctl sources in a pkg.devel repository

@nickboldt why should we continue maintaining a `deploy.sh` script? The goal was to replace deploy.sh with crwctl. I am confused.

@ibuziuk @nickboldt chectl build is [currently cloning](https://github.com/che-incubator/chectl/blob/master/package.json#L30) https://github.com/eclipse/che-operator/ to get CR and deploy files. crwctl build should clone a different repository (https://github.com/redhat-developer/codeready-workspaces) and we should move crw CR and deploy files (extracted from deploy.sh) there. Does that make sense?

nickboldt:
> @nickboldt why should we continue maintaining a deploy.sh script? The goal was to replace deploy.sh with crwctl. I am confused.

As discussed today if crwctl embeds the cr.yaml and other defaults (`operator-rhel8:2.0` image) then there's no need for a wrapper as the wrapper would only be doing:

`crwctl server:start`

But if we need to set various flags, like:

```
crwctl server:start -a operator -p openshift --domain=${OPENSHIFT_CONSOLE_URL} \
  --che-operator-image=quay.io/crw/operator-rhel8:2.0 \
  --che-operator-cr-yaml=custom-resource.yaml -n ${OPENSHIFT_PROJECT}
```

Then there might be value in making a wrapper to collect logical defaults into one place. Otherwise, we just make sure the commonly needed flags are documented fully.

cc: @rkratky

l0rd:
> ```
> crwctl server:start -a operator -p openshift --domain=${OPENSHIFT_CONSOLE_URL} \
>   --che-operator-image=quay.io/crw/operator-rhel8:2.0 \
>   --che-operator-cr-yaml=custom-resource.yaml -n c
> ```

In `crwctl` options `-a` and `-p` don't make sense and the rest should be the default. 

Oh and another thing: in `chectl` you don't need to `-n ${OPENSHIFT_PROJECT}`, if you export the variable `${CHE_NAMESPACE}` then you don't need to specify `-n` when you invoke `chectl server:start`, the namespace will be set automatically (that's set [here](https://github.com/che-incubator/chectl/blob/master/src/common-flags.ts#L16)).

nickboldt:
> CHE_NAMESPACE

a) it's too bad that we're still naming things with "Che" when we KNOW we plan to whitelabel/productize. PPROJECT_NAMESPACE would have been better.

b) if step 1 of using crwctl is "export a variable" then it's not as simple as `crwctl server:start` which means again.. a wrapper script might be useful to capture all the documented steps. 

I mean you, as an end user... would YOU want to have to refer to the doc every time you want to deploy something? or would you rather just run a script that's already in the tarball from which you got crwctl?

l0rd:
> a) it's too bad that we're still naming things with "Che" when we KNOW we plan to whitelabel/productize. PPROJECT_NAMESPACE would have been better.

Prefixing the environment varibles with `CHE_` has been a convention that we have set a couple of years ago to [automatically pass those variables to the Kubernetes deployment](https://github.com/eclipse/che/blob/master/deploy/openshift/deploy_che.sh#L66). That's quite helpful but has the downside to require extra-work for whitelabelling Che. I am +1 to start a discussion to improve that.

> b) if step 1 of using crwctl is "export a variable" then it's not as simple as crwctl server:start which means again.. a wrapper script might be useful to capture all the documented steps.

No I don't mean that. It works as for deploy.sh: if the variable is not exported by the user it will use the default value. If a user wants to change it he needs to export the variable.

themr0c:
We will need some clear information for the downstream Installation Guide:

- Can we safely replace all mentions to `chectl` by  mentions to the new name?
- Is `crwctl` the definitive name for downstream `chectl`?
- Download URL and installation procedure? How much can we reuse?

nickboldt:
> Can we safely replace all mentions to chectl by mentions to the new name? 

Yes, make `chectl` a parameterized variable so you can replace it in Che 7 docs w/ `crwctl` in CRW 2 docs.

> Is crwctl the definitive name for downstream chectl?

At this point, yes. Derived from kubectl -> chectl -> crwctl.

> Download URL and installation procedure? How much can we reuse?

I expect we'll have a tarball including at least the crwctl binary + readme that will be similar to the access.redhat.com docs. Take the URL for 1.2.x.GA and replace that with 2.0.0.GA and you should be set.

"installation" of crwctl is "unpack the tarball, then run `crwctl server:start ...`. "

Obviously there's more to it -- will have to list the commonly used flags for the server:start command. For 2.1 (or between 2.0 and 2.1) we can add more detail about other commands.

ibuziuk:
closing in favour of https://github.com/redhat-developer/codeready-workspaces-chectl/pull/1

PR #381, opened 2019-11-05
[WIP] Skip creating the clusterrole/clusterrolebinding if we are not enabling openshif OAuth

tomgeorge:
Signed-off-by: Tom George <tg82490@gmail.com>

### What does this PR do?
Skip creating the `che-operator` clusterrole if `--os-auth` is not enabled.

Part of https://github.com/eclipse/che/issues/14662 that I changed as part of my testing.

### DIFF
--- a/src/tasks/installers/operator.ts
+++ b/src/tasks/installers/operator.ts
@@ -91,6 +91,11 @@ export class OperatorTasks {
       },
       {
         title: `Create ClusterRole ${this.operatorClusterRole}`,
+        skip: () => {
+          if (!flags['os-oauth']) {
+            return `No need for ClusterRole ${this.operatorClusterRole}`
+          }
+        },
         task: async (_ctx: any, task: any) => {
           const exist = await kube.clusterRoleExist(this.operatorClusterRole)
           if (exist) {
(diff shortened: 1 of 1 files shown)


src/tasks/installers/operator.ts
```diff
@@ -120,6 +125,11 @@ export class OperatorTasks {
       },
       {
         title: `Create ClusterRoleBinding ${this.operatorClusterRoleBinding}`,
+        skip: () => {
+          if (!flags['os-oauth']) {
+            return `No need for ClusterRole ${this.operatorClusterRole}`
```
sleshchenko: Previously it was implemented in such way, but then I removed such check during console link support implementing.

How consolelink supposed to work after your changes? https://github.com/eclipse/che-operator/blob/master/deploy/cluster_role.yaml#L41

sleshchenko: https://github.com/che-incubator/chectl/pull/291

sleshchenko: I think the solution should be smarter - chectl can check is the current user has admin rights:
- if has - create cluster role, everything is OK
- if it does not - OAuth flag is not permitted, during deploying chectl prints warning that consolelink won't be created.

It's needed to check what che-operator will do on OpenShift 4.2 when he does not permissions to create console link, probably it fails to deploy Che...

tomgeorge:
Would you say this is better suited as an operator change, to separate those clusterroles?

Then on the chectl side, if `os-oauth` skip creation of those roles?

PR #383, opened 2019-11-07
feat: CheCtl start should *not* fix the docker images in the CR by default

tomgeorge:
### What does this PR do?

Modifies the logic for operator installation when assigning tags to images:

1. If `--cheimage` is not overriden on the CLI, and the user is running a release version of `chectl`, it will set the che, devfile, plugin registry, and keycloak images to the empty string, in order to use the operator defaults
2. If a user is running a `nightly` or `latest` version of `chectl`, or the user overrides the che image with  `--cheimage` CLI argument, chectl will overwrite the docker image tags in the custom resource to match the che image tag.

https://github.com/eclipse/che/issues/15131


ibuziuk:
@davidfestal should we also backport it to 7.3.x ?

davidfestal:
> @davidfestal should we also backport it to 7.3.x ?

I think so yes

PR #394, opened 2019-11-13
fix: Fix server update for CRW

davidfestal:
#384 # What does this PR do?

This PR fixes a number of issues in the chectl code that still prevent it to be used for installing the CRW flavor:
- Replace hard-coded value of the `che-operator` name by the corresponding variables, in order to allow using alternate values in the CRW case
- Allow deleting / creating again a deployment when it cannot be directly updated due to immutable fields

This references the following JIRA issue comment: https://issues.jboss.org/browse/CRW-462?focusedCommentId=13811780&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13811780

### DIFF
--- a/src/api/kube.ts
+++ b/src/api/kube.ts
@@ -775,7 +775,7 @@ export class KubeHelper {
     try {
       return await k8sAppsApi.replaceNamespacedDeployment(yamlDeployment.metadata.name, namespace, yamlDeployment)
     } catch (e) {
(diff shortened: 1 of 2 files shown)


ibuziuk:
Could we backport it to 7.3.x ?

davidfestal:
> Could we backport it to 7.3.x ?

We should yes. Maybe after having squashed/merged this one on master, no ?

PR #402, opened 2019-11-16
CRW-483: The 'server:start' command of the 'oc' client fails on the OCP 3.11' cluster with state 'Running' / 'Pending' expected

ibuziuk:
### What does this PR do?
CRW-483: The 'server:start' command of the 'oc' client fails on the OCP 3.11' cluster with state 'Running' / 'Pending' expected

https://issues.jboss.org/browse/CRW-483


benoitf:
@ibuziuk FYI semantic check is failing

Commit message should start with for example 
`fix(serveur:start): blahblahblah`

PR #412, opened 2019-11-21
fix: Increase microk8s.status timeout from 1s to 10s

NighttimeDriver50000:
`microk8s.status` can sometimes take longer than 1 second to execute, causing chectl to erroneously determine that it is not running.

### What does this PR do?
This PR increases the timeout on the async call to microk8s.status from 1s to 10s.

I did not create an issue, as the change is very small and simple. This PR does, however, fix an issue I had on my machine with microk8s.status taking more than 1s to run.


benoitf:
thx @NighttimeDriver50000

PR #417, opened 2019-11-25
feat(helm): Add support for helm v3

benoitf:
### What does this PR do?
feat(helm): Add support for helm v3

Fixes https://github.com/eclipse/che/issues/15235

I tested with docker-desktop

```
$ chectl server:start --platform=docker-desktop --installer=helm
  ✔ Verify Kubernetes API...OK
  ✔ 👀  Looking for an already existing Che instance
    ✔ Verify if Che is deployed into namespace "che"...it is not
  ✔ ✈️  Docker Desktop preflight checklist
    ✔ Verify if kubectl is installed
    ✔ Verify if kubectl context is Docker Desktop: Found docker-desktop.
    ✔ Verify remote kubernetes status...done.
    ✔ Verify if nginx ingress is installed
    ↓ Installing nginx ingress [skipped]
      → Ngninx ingress is already setup.
    ✔ Verify domain is set... auto-assigning domain to 192.168.1.11.nip.io....set to 192.168.1.11.nip.io.
  ✔ 🏃‍  Running Helm to install Che
    ✔ Verify if helm is installed
    ✔ Check Helm Version: Found v3.0.0+ge29ce2a
    ✔ Create Namespace (che)...done.
    ✔ Check Cluster Role Binding...does not exists.
    ✔ Preparing Che Helm Chart...done.
    ✔ Updating Helm Chart dependencies...done.
    ✔ Deploying Che Helm Chart...done.
  ✔ ✅  Post installation checklist
    ✔ Devfile registry pod bootstrap
      ✔ scheduling...done.
      ✔ downloading images...done.
      ✔ starting...done.
    ✔ Plugin registry pod bootstrap
      ✔ scheduling...done.
      ✔ downloading images...done.
      ✔ starting...done.
    ✔ Che pod bootstrap
      ✔ scheduling...done.
      ✔ downloading images...done.
      ✔ starting...done.
    ✔ Retrieving Che Server URL...http://che-che.192.168.1.11.nip.io
    ✔ Che status check
Command server:start has completed successfully.
```

note: there is another commit that enable back some unit tests

### DIFF
--- a/src/tasks/installers/helm.ts
+++ b/src/tasks/installers/helm.ts
@@ -16,6 +16,7 @@ import { copy, mkdirp, remove } from 'fs-extra'
 import * as Listr from 'listr'
 import * as path from 'path'
 
+import { CheHelper } from '../../api/che'
 import { KubeHelper } from '../../api/kube'
 
 export class HelmTasks {
@@ -27,6 +28,33 @@ export class HelmTasks {
       {
         title: 'Verify if helm is installed',
         task: () => { if (!commandExists.sync('helm')) { command.error('E_REQUISITE_NOT_FOUND') } }
+      },
+      {
+        title: 'Check Helm Version',
+        task: async (ctx: any, task: any) => {
+          try {
+            const version = await this.getVersion()
+            if (version.startsWith('v3.')) {
+              ctx.isHelmV3 = true
+            }
+            task.title = await `${task.title}: Found ${version}`
+          } catch (error) {
+            command.error(`Unable to get helm version. ${error.message}`)
+          }
+        }
+      },
+      {
+        title: `Create Namespace (${flags.chenamespace})`,
+        task: async (_ctx: any, task: any) => {
+          const che = new CheHelper(flags)
+          const exist = await che.cheNamespaceExist(flags.chenamespace)
+          if (exist) {
+            task.title = `${task.title}...does already exist.`
+          } else {
+            await execa(`kubectl create namespace ${flags.chenamespace}`, { shell: true })
+            task.title = `${task.title}...done.`
+          }
+        }
       },
       {
         title: 'Check for TLS secret prerequisites',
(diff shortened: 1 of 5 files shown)


src/tasks/installers/helm.ts
```diff
@@ -28,6 +29,33 @@ export class HelmTasks {
...
+          }
+        }
+      },
+      {
+        title: `Create Namespace (${flags.chenamespace})`,
+        task: async (_ctx: any, task: any) => {
+          const che = new CheHelper(flags)
+          const exist = await che.cheNamespaceExist(flags.chenamespace)
+          if (exist) {
+            task.title = `${task.title}...does already exists.`
```
sleshchenko: I'm not sure but maybe

```suggestion
            task.title = `${task.title}...does already exist.`
```

src/tasks/installers/helm.ts
```diff
@@ -96,9 +142,13 @@ export class HelmTasks {
       },
       {
         title: 'Create Tiller RBAC',
+        // Tiller is not used anymore in helm v3
+        enabled: (ctx: any) => !ctx.isHelmV3,
```
sleshchenko: It may be easier to read and understand if v3 and !v3 tasks will be in different methods.

I mean something like:
```
[{check version},
 { 
  if v3 
  getV3Tasks
 else 
  get!V3Tasks
}]
```

benoitf: If I do this it will do subtree in the output

src/tasks/installers/helm.ts
```diff
@@ -82,8 +112,24 @@ export class HelmTasks {
           }
         }
       },
+      {
+        title: 'Check Cluster Role Binding',
```
l0rd: Do we really need to remove the cluster role binding if it exist? If it's not mandatory I would not add this extra step because it may fail (for example because the user doesn't have enough privileges), it's something the user hasn't requested and it doesn't bring any value.

benoitf: Issue is that helm complain about existing role if I launch server start again if I remove che namespace

benoitf: Issue is that helm complain about existing role if I launch server start again if I remove che namespace

src/tasks/installers/helm.ts
sleshchenko: Ok, then I'm ok to merge it as is

benoitf:
I created the issue there : https://github.com/eclipse/che/issues/15333

PR #450, opened 2020-01-02
fix: Store server logs into system tmp directory by default

tolusha:

### What does this PR do?
This PR allows to store Eclipse Che logs into a system tmp directory by default


sparkoo:
great, thanks!

PR #465, opened 2020-01-15
feat: Collect k8s events

tolusha:

### What does this PR do?
Collect k8s events as well as che logs.

The project structure:
```
.
├── events.txt
├── mkdir-workspacet93uo90gd1acw8zj
│   └── mkdir-workspacet93uo90gd1acw8zj.log
├── workspacet93uo90gd1acw8zj.che-jwtproxy-6454d8877b-5xp26
│   └── verifier.log
└── workspacet93uo90gd1acw8zj.go-cli-bf5459458-fgrdx
    ├── che-machine-execrbr.log
    ├── go-cli.log
    ├── theia-ide3j0.log
    └── vscode-goqsw.log

3 directories, 7 files
```

https://github.com/eclipse/che/issues/15604

### DIFF
--- a/src/commands/server/start.ts
+++ b/src/commands/server/start.ts
@@ -238,7 +238,7 @@ export default class Start extends Command {
 
     const eventTasks = new Listr([{
       title: 'Start following events',
-      task: () => new Listr(cheTasks.namespaceEventsTask(flags, this, true))
+      task: () => new Listr(cheTasks.namespaceEventsTask(flags.chenamespace, this, true))
     }], listrOptions)
 
     try {
--- a/src/commands/server/start.ts
+++ b/src/commands/server/start.ts
@@ -112,7 +112,8 @@ export default class Start extends Command {
     }),
     directory: string({
       char: 'd',
-      description: 'Directory to store logs into'
+      description: 'Directory to store logs into',
+      env: 'CHE_LOGS'
     }),
     'workspace-pvc-storage-class-name': string({
       description: 'persistent volume(s) storage class name to use to store Eclipse Che workspaces data',
@@ -235,13 +236,20 @@ export default class Start extends Command {
       task: () => new Listr(cheTasks.serverLogsTasks(flags, true))
     }], listrOptions)
 
+    const eventTasks = new Listr([{
+      title: 'Start following events',
+      task: () => new Listr(cheTasks.namespaceEventsTask(flags, this, true))
+    }], listrOptions)
+
     try {
       await preInstallTasks.run(ctx)
 
       if (!ctx.isCheDeployed) {
         this.checkPlatformCompatibility(flags)
         await platformCheckTasks.run(ctx)
+        this.log(`Eclipse Che logs will be available in '${ctx.directory}'`)
         await logsTasks.run(ctx)
+        await eventTasks.run(ctx)
         await installTasks.run(ctx)
       } else if (!ctx.isCheReady
         || (ctx.isPostgresDeployed && !ctx.isPostgresReady)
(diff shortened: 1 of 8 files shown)


sleshchenko:
@tolusha Seems this PR has conflicts against master branch.
Also, it could be useful if you add examples of folder structure we got as a result, along with file content examples (hint probably some info may be copied/pasted from the referenced issue description but I'm not sure if it's aligned to the implementation you provided).

src/commands/server/start.ts
```diff
@@ -226,13 +227,20 @@ export default class Start extends Command {
       task: () => new Listr(cheTasks.serverLogsTasks(flags, true))
     }], listrOptions)
 
+    const eventTasks = new Listr([{
+      title: 'Start following events',
+      task: () => new Listr(cheTasks.namespaceEventsTask(flags, this, true))
+    }], listrOptions)
+
     try {
+      this.log(`Eclipse Che logs will be available in '${ctx.directory}'`)
```
sleshchenko: I faced some inconsistent behavior and planned to register an issue for that, but it seems it should be easy to solve and can be done here.
So, I think it would be better to print such an output after a platform checks tasks will be done,
otherwise user will be notified that Che Logs will be available but actually there is not chance for this to happen since user did not specified platform and there will not be even try to deploy Che

yarn.lock
```diff
@@ -1677,11 +1677,11 @@ ecc-jsbn@~0.1.1:
 
 "eclipse-che-operator@git://github.com/eclipse/che-operator#master":
   version "0.0.0"
-  resolved "git://github.com/eclipse/che-operator#2e47c3331782868d7b120718ded7be5f5e64c9da"
+  resolved "git://github.com/eclipse/che-operator#603f14a3d9c12b5c732870333787ec2748095ebb"
```
sleshchenko: are you sure that you need to update it within this PR?

src/commands/workspace/logs.ts
```diff
@@ -47,21 +48,23 @@ export default class Logs extends Command {
     const cheTasks = new CheTasks(flags)
     const k8sTasks = new K8sTasks()
 
+    const follow = flags.follow
```
sleshchenko: it's personal taste. But I would not declare here const and use directly flags.follow as before.

src/commands/workspace/logs.ts
```diff
@@ -47,21 +48,23 @@ export default class Logs extends Command {
...
     const tasks = new Listr([], { renderer: flags['listr-renderer'] as any })
     tasks.add(k8sTasks.testApiTasks(flags, this))
     tasks.add(cheTasks.verifyCheNamespaceExistsTask(flags, this))
-    if (!flags.follow) {
+    if (!follow) {
       tasks.add(cheTasks.verifyWorkspaceRunTask(flags, this))
     }
     tasks.add(cheTasks.workspaceLogsTasks(flags))
+    tasks.add(cheTasks.namespaceEventsTask(flags, this, follow))
```
sleshchenko: Well, it will listen to events in a namespace where Che Server is deployed and it's not really thing we should do...
Maybe instead it would be better to collects events in namespace where workspace is really started (default value now is `<username>-che`). You can fetch this info from Che Server/api/workspace/{workspaceid}#attributes.infrastructureNamespace.

Please also consider filter objects by workspaceId label.

yarn.lock
tolusha: I don't see anything bad in depending on the latest

sleshchenko: I just don't understand what is a process of updating dependencies in yarn.lock, it's why I'm asking.

tolusha:
@sleshchenko It is ready to review

PR #476, opened 2020-01-30
feat: Implement TLS by default for Minikube + Helm installer

mmorhun:

### What does this PR do?
Implements automatic installation of Che in TLS mode for Minikube infrastructure and Helm installer.
Theoretically should also work on Kubernetes and MicroK8S

### How to test
To test, build this PR with changes from https://github.com/eclipse/che/pull/15878 into Helm templates (just change in chectl's `package.json` `eclipse-che` dependency to point to the `che-15810` branch).
Then, just run `chectl server:start --platform=minikube --multiuser --installer=helm`
To install single-user Che, omit `--multiuser` flag.

### How it works
At the checks stage chectl checks for `che-tls` secret presence. If it is not present, then the procedure of auto-generation begins. If the secret is pre-created (current scenario) then nothing is changes and everything should work as it was before.
In case of auto-generation, first, chectl installs Cert Manager (using the pre-cached [yaml](https://github.com/jetstack/cert-manager/releases/download/v0.13.0/cert-manager.yaml) from official site) for managing tls secrets for Che. This involves generation of CA credentials for Cert Manager, which is done via dedicated job. When all requirements are satisfied, chectl requests from Cert Manager tls secret for Che encryption.
The only manual step which is required from end users is to add CA certificate into their browsers.

https://github.com/eclipse/che/issues/15313
Should be merged together with https://github.com/eclipse/che/pull/15878


resources/cert-manager/che-cluster-issuer.yml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: cert-manager.io/v1alpha2
+kind: ClusterIssuer
+metadata:
+    name: che-cluster-issuer
+    namespace: cert-manager
+spec:
+   ca:
```
sleshchenko: This file seems not to be well formatted with spaces.

src/commands/server/start.ts
```diff
@@ -200,6 +217,13 @@ export default class Start extends Command {
     const ctx: any = {}
     ctx.directory = path.resolve(flags.directory ? flags.directory : path.resolve(os.tmpdir(), 'chectl-logs', Date.now().toString()))
     const listrOptions: Listr.ListrOptions = { renderer: (flags['listr-renderer'] as any), collapse: false, showSubtasks: true } as Listr.ListrOptions
+    ctx.listrOptions = listrOptions
+
+    // TODO temporary workaround.
+    // When tls by default is implemented for all plutforms, delete `tls` flag
```
sleshchenko: plutforms -> platforms

src/constants.ts
```diff
@@ -1,4 +1,16 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ **********************************************************************/
 
-export { DEFAULT_CHE_IMAGE, DEFAULT_CHE_OPERATOR_IMAGE }
+export const DEFAULT_CHE_IMAGE = 'eclipse/che-server:nightly'
+export const DEFAULT_CHE_OPERATOR_IMAGE = 'quay.io/eclipse/che-operator:nightly'
+
+export const CERT_MANAGER_NAMESPACE_NAME = 'cert-manager'
+
+export const CHE_SECRET_NAME = 'che-tls'
```
sleshchenko: It's tls specific secret, right? Then let's make it clearer with constant name

README.md
```diff
@@ -243,24 +243,24 @@ OPTIONS
...
 
   -p, --platform=minikube|minishift|k8s|openshift|microk8s|docker-desktop|crc
-      Type of Kubernetes platform. Valid values are "minikube", "minishift", "k8s (for kubernetes)", "openshift", "crc 
+      Type of Kubernetes platform. Valid values are "minikube", "minishift", "k8s (for kubernetes)", "openshift", "crc
       (for CodeReady Containers)", "microk8s".
 
   -s, --tls
       Enable TLS encryption.
-                           Note that for kubernetes 'che-tls' with TLS certificate must be created in the configured 
+                           Note that for kubernetes 'che-tls' with TLS certificate must be created in the configured
```
sleshchenko: It makes sense to add a note that TLS now is enabled for k8s, minikube... platforms and there is no way to disable it.
Well, maybe it would be better to leave this option for all platforms but with default option true? I mean keeping an ability to run `server:start --tls=false`

src/commands/server/start.ts
```diff
@@ -43,6 +43,11 @@ export default class Start extends Command {
       default: Start.getTemplatesDir(),
       env: 'CHE_TEMPLATES_FOLDER'
     }),
+    resources: string({
+      description: 'Path to the resources folder',
```
sleshchenko: What are resources and templates folders for chectl... It's not clear for user how to use them I believe
Che Templates is out of scope the current PR but since you introduce resources folder, I would like to think more about UX here.

src/tasks/component-installers/cert-manager.ts
```diff
@@ -0,0 +1,184 @@
...
+      },
+      {
+        title: 'Request self-signed certificate',
+        task: async (ctx: any, task: any) => {
+          if (ctx.cheCertificateExists) {
+            throw new Error('Che certificate already exists.')
+          }
+
+          const certificateTemplatePath = path.join(flags.resources, '/cert-manager/che-certificate.yml')
+          await this.kubeHelper.createCheClusterCertificate(certificateTemplatePath, flags.domain)
```
sleshchenko: I tried to deploy secure Che and reached timeout first time
Screenshot_20200131_092437

I suppose it's because time is consumed by cert manager... I wonder if there is global timeout...

Then I tried to redo the same command and got the following
Screenshot_20200131_092452


Please take a look issues I faced and consider improving UX here

sleshchenko:
`rm -rf ~/.cache/chectl` helped to solve my issue with failed Che Server :tada:

src/tasks/component-installers/cert-manager.ts
```diff
@@ -0,0 +1,184 @@
...
+      {
+        title: 'Add local Che CA certificate into browser',
+        task: async (_ctx: any, task: any) => {
+          const cheSecret = await this.kubeHelper.getSecret(CHE_SECRET_NAME, flags.chenamespace)
+          if (cheSecret && cheSecret.data) {
+            const cheCaCrt = Buffer.from(cheSecret.data['ca.crt'], 'base64').toString('ascii')
+            const cheCaPublicCertPath = path.join(os.homedir(), 'cheCA.crt')
+            fs.writeFileSync(cheCaPublicCertPath, cheCaCrt)
+
+            task.title = `❗[MANUAL ACTION REQUIRED] Please add local Che CA certificate into your browser: ${cheCaPublicCertPath}`
```
sleshchenko: Can we make MANUAL ACTION REQUIRED like colored to yellow?
Screenshot_20200131_102126

Firefox allows me to import that file and it works like a charm!

But with the latest stable Chrome (Version 79.0.3945.130 (Official Build) (64-bit)) I got the following error
Screenshot_20200131_102202

resources/cert-manager/che-cluster-issuer.yml
mmorhun: Yes, thanks, will align it correctly

src/constants.ts
mmorhun: ok

src/tasks/component-installers/cert-manager.ts
mmorhun: Yes, will take a look if we can do something around coloring

mmorhun: I have exactly the same version of Gooogle Chrome: 
image
And everything works like a charm:
image

mmorhun: @sleshchenko can you please describe step-by-step how exactly your add the CA certificate into your Google Chrome browser?

README.md
mmorhun: My plan was to remove that option at all when TLS by default is finished. In most cases people use default Che-Theia editor which in some cases doesn't work properly without TLS  (all webview are broken). And, of course, the security...
We may leave the flag, but make it default to `true`, I do not mind. However, in such case we should test non-TLS installations as well... I think this should be decided on wide level.

As about note in readme, ok, will write a note.

benoitf: maybe keep it for backward compliance (but just to add deprecation notice when using it saying that now Eclipse Che is always in secured mode)

src/tasks/component-installers/cert-manager.ts
mmorhun: Thank you for letting me know. The problem is not in the timeout. I missed one place where namespace `che` was hardcoded. I've fixed that and now it can deploy Che in custom namespace as well.

src/commands/server/start.ts
mmorhun: Yes, that's not clear, but it is the same for files under templates folder. I can move cert-manager yaml files into templates folder, but I think that for CA certificate generation job it should be another location.

tolusha:
@mmorhun 
README.md is automatically generated from sources when you launch `yarn pack`

README.md
tolusha: Agree with @benoitf

src/commands/server/start.ts
tolusha: @mmorhun 
What do you want to store under `resources` folder?

src/tasks/component-installers/cert-manager.ts
```diff
@@ -0,0 +1,184 @@
...
+        title: 'Check Cert Manager deployment',
+        task: async (ctx: any, task: any) => {
+          // Check only one CRD of cert-manager assuming that it is installed or not.
+          ctx.certManagerInstalled = await this.kubeHelper.namespaceExist(CERT_MANAGER_NAMESPACE_NAME) && await this.kubeHelper.crdExist('certificates.cert-manager.io')
+          if (ctx.certManagerInstalled) {
+            task.title = `${task.title}...already deployed`
+          } else {
+            task.title = `${task.title}...not deployed`
+
+            return new Listr([
```
tolusha: It is not good to return tasks inside task.
It is better to skip unnecessary using `enable` property. And the flag can be put into ctx

src/commands/server/start.ts
mmorhun: @tolusha I've moved content of `resources` folder into `templates` folder.

README.md
mmorhun: Ok, will do that.

src/tasks/component-installers/cert-manager.ts
mmorhun: @tolusha I think it is a standard way to go. We may organize the sequence of tasks in more clear way: create a separate method which returns subtasks list. And that it will be clear to read (more we'll have displaying of the tasks as subtasks).
But if you really want to have flattened list of tasks, let it be so...

mmorhun:
@tolusha thank you, I didn't know that readme is autogenerated... Fix move changes to the source code.

README.md
```diff
@@ -231,7 +231,7 @@ OPTIONS
       show CLI help
 
   -i, --cheimage=cheimage
-      [default: quay.io/eclipse/che-server:nightly] Che server container image
+      [default: eclipse/che-server:nightly] Che server container image
```
tolusha: This seems strange. 
Should be `quay.io`

tomgeorge:
At a high-level, I think it would be good to add a flag for the cert-manager namespace, in case the user already has their own cert-manager deployed that they want to use, in a namespace other than `cert-manager`.  

Without knowing the context that went into deciding on using cert-manager for this, pulling in another cluster-addon component seems a little heavyweight when creating a TLS secret is all we need to do.

mmorhun:
@tomgeorge thanks for the idea, but having two Cert Managers installed in the cluster would be weird. I think, instead we should request the secret from already existed Cert Manager, if any.
But that is out of scope of this PR.

tomgeorge:
@mmorhun sorry, I meant that if a cert-manager is already installed, che should use that one.  But understand that it is out of scope for this PR.

README.md
```diff
@@ -270,8 +270,9 @@ OPTIONS
 
   -s, --tls
       Enable TLS encryption.
-                           Note that for kubernetes 'che-tls' with TLS certificate must be created in the configured 
-      namespace.
+                           Note, that this option is turned on by default for kubernetes infrastructure.
+                           If it is needed to provide own certificate, 'che-tls' secret with TLS certificate must be
```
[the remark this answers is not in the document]
mmorhun: I see no mistakes in wording here. Maybe 'if user wants' will be clearer for you, but it is not docs style, I think.

README.md
```diff
@@ -313,8 +314,9 @@ OPTIONS
       persistent volume storage class name to use to store Eclipse Che Postgres database
 
   --self-signed-cert
-      Authorize usage of self signed certificates for encryption. Note that `self-signed-cert` secret with CA certificate 
-      must be created in the configured namespace.
+      Authorize usage of self signed certificates for encryption.
+                           This is the flag for Che to propagate the certificate to components, so they will trust it.
+                           Note that `che-tls` secret with CA certificate must be created in the configured namespace.
```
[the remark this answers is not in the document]
mmorhun: Not completely. Think of case when user wants to provide own certificate.

src/tasks/component-installers/cert-manager.ts
```diff
@@ -0,0 +1,182 @@
...
+          // To be able to use self-signed sertificate it is required to provide CA private key & certificate to cert-manager
+          const caSelfSignedCertSecret = await this.kubeHelper.getSecret(CERT_MANAGER_CA_SECRET_NAME, CERT_MANAGER_NAMESPACE_NAME)
+          if (!caSelfSignedCertSecret) {
+            // First run, generate CA self-signed certificate
+
+            task.title = `${task.title}...generating new one`
+
+            const CA_CERT_GENERATION_SERVICE_ACCOUNT_NAME = 'ca-cert-generator'
+            const CA_CERT_GENERATION_JOB_NAME = 'ca-cert-generation-job'
+            const CA_CERT_GENERATION_JOB_IMAGE = 'quay.io/eclipse/che-cert-manager-ca-cert-generator:latest'
```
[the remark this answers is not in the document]
mmorhun: Yes, thants the plan for next PR. I want to move dockerfile into Che repository and automate the build.

templates/cert-manager/che-cluster-issuer.yml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: cert-manager.io/v1alpha2
+kind: ClusterIssuer
+metadata:
+    name: che-cluster-issuer
```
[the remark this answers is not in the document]
mmorhun: Ok, will fix it.

mmorhun:
@tomgeorge sorry, I probably misunderstood you. Usually, Cert Manager is installed in `cert-manager` namespace. For some installation methods (for example from yaml file) it is even hardcoded.

src/api/kube.ts
```diff
@@ -8,17 +8,22 @@
...
 import { cli } from 'cli-ux'
+import * as execa from 'execa'
 import * as fs from 'fs'
 import https = require('https')
 import * as yaml from 'js-yaml'
 import { Writable } from 'stream'
 
 import { DEFAULT_CHE_IMAGE } from '../constants'
 
+import { V1alpha2Sertificate } from './typings/cert-manager'
```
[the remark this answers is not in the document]
mmorhun: No, typo, thank you for letting me know. Definitely will fix.

mmorhun:
@tomgeorge I've fixed all your remarks. Thanks for spotting.

mmorhun:
@tomgeorge could you please take a look at this PR again and add more comments or unblock the merge?

README.md
```diff
@@ -270,8 +270,9 @@ OPTIONS
 
   -s, --tls
       Enable TLS encryption.
-                           Note that for kubernetes 'che-tls' with TLS certificate must be created in the configured 
-      namespace.
+                           Note, that this option is turned on by default for kubernetes infrastructure.
```
metlos: So when this is on by default, how do you actually switch it off? Do we have something like `--no-tls`?

mmorhun: No we don't have `--no-tls` flag, so a user cannot switch it off on Kubernetes-like infrastructures. The same is planned for Openshift soon. Although, we are going to keep the `--tls` flag for backward compatibility only.
Turning it off not only dangerous from user security point of view, but also some components might not work properly without TLS (as you probably know the story about Theia webviews).

templates/cert-manager/cert-manager.yml
```diff
@@ -0,0 +1,6409 @@
+# This file is copied from:
```
metlos: How are we going to support this file in CRW? Have we checked that the licenses are compatible? https://github.com/jetstack/cert-manager/blob/master/LICENSES

Also, all these definitions (apart from the CRDs) are managed through helm. Would it not be possible to use helm for installing them instead of hardcoding this humongous file into our codebase?

README.md
nickboldt: so if TLS is on by default and no longer disableable, the flag would effectively do nothing? 

Will you also ensure the fields in custom resource.yaml no longer work, so that it's no longer possible to set `tlsSupport: false` ?

dockerfiles/cert-manager-ca-cert-generator-job/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# CA key pair generation job container
+FROM alpine
+
+RUN apk add --no-cache openssl curl && \
+    cd /usr/local/bin && \
+    curl -s -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl && \
```
nickboldt: seems strange to join the `-L` and `-O` flags but not the `-s` flag here

src/tasks/component-installers/cert-manager.ts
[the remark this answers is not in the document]
nickboldt: Could this be based on UBI instead of Alpine to make it a bit easier to productize? That way you could potentially RUN this build on demand rather than having Yet Another Container To Build, Publish, Maintain, Keep CVE Free, and Productize.

dockerfiles/cert-manager-ca-cert-generator-job/entrypoint.sh
```diff
@@ -0,0 +1,36 @@
...
+            -outform PEM -out $CA_CERT_FILE
+            # Do not include CA:TRUE as it is already included into default config file
+            #-addext basicConstraints=critical,CA:TRUE
+
+# Create CA root certificate secret
+
+CERT_MANAGER_NAMESPACE='cert-manager'
+CERT_MANAGER_CA_SECRET_NAME='ca'
+
+kubectl create secret tls $CERT_MANAGER_CA_SECRET_NAME --key=$CA_KEY_FILE --cert=$CA_CERT_FILE --namespace $CERT_MANAGER_NAMESPACE
```
nickboldt: So.. .this container is just kubectl + a cert? Instead of new container, could we instead just add this cert into https://github.com/che-dockerfiles/che-sidecar-kubernetes-tooling and keep the proliferation of containers (and therefore things that need to be productized) to a minimum? CRW 1.2 was 8 containers. 2.0 was 20. 2.1 is up to 23 and counting. Can we try to reuse out existing tiny containers and just add a new kb of new files into them??

README.md
mmorhun: @nickboldt yes the flag will do nothing... Please see the thread https://github.com/che-incubator/chectl/pull/476#discussion_r373383036.

mmorhun: As about `tlsSupport` flag in CRDs... you are right, we have to take a look and prevent breaking stuff with wrong value of the flag.

templates/cert-manager/cert-manager.yml
mmorhun: > Have we checked that the licenses are compatible? 

Cert Manager [is licensed](https://github.com/jetstack/cert-manager/blob/master/LICENSES#L2) under Apache License Version 2.0. That mean we can use it.

> Would it not be possible to use helm for installing them

Yes, it is possible. But we plan to use the same approach for operator. And there using Helm will be weird.

dockerfiles/cert-manager-ca-cert-generator-job/entrypoint.sh
mmorhun: @nickboldt yes, this container indeed contains kubectl and openssl only, but we cannot reuse Kubernetes tolling sidecar for the Cert Manager job. First, they have different purposes: the job completes its tasks and exists, whereas Kubernetes tooling sidecar is designed to be run as a part of a workspace. Second, they have different aims: generate certificate and serve queries to Kubernetes respectively. Third, chectl (and Che installation process) should not depend on a Che sidecar. That's logically wrong and error prone: if something is changed in Kubernetes Che plugin it could break Che installation.

README.md
mmorhun: But I think it should be done when I'll be working on operator part.

nickboldt: OK as long as this TODO is captured in a GH issue for that task, then +1. Just don't want to forget this part. :D

dockerfiles/cert-manager-ca-cert-generator-job/entrypoint.sh
nickboldt: *sigh* OK, I was hoping for reuse and less proliferation of single-use containers. Your logic is sound, but still makes me sad. :(

src/tasks/component-installers/cert-manager.ts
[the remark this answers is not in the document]
nickboldt: I guess we can't build on demand but moving to UBI would still make downstream easier. Or else I can contribute a rhel.Dockerfile variant that we can use in CRW.

mmorhun:
@rhopp FYI: we are going to merge this PR right after 7.9 release. So, if you have any infrastructure which uses Helm Che installer, from that point, it will deploy Che in TLS mode with self-signed certificate.

rhopp:
👍

PR #477, opened 2020-01-30
feat: Allow to debug Eclipse Che Server locally

tolusha:

### What does this PR do?
Introduce `server:debug` commands with is basically starts forward port to debug Eclipse Che server locally.

https://github.com/eclipse/che/issues/14976

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ USAGE
 * [`chectl autocomplete [SHELL]`](#chectl-autocomplete-shell)
 * [`chectl devfile:generate`](#chectl-devfilegenerate)
 * [`chectl help [COMMAND]`](#chectl-help-command)
+* [`chectl server:debug`](#chectl-serverdebug)
 * [`chectl server:delete`](#chectl-serverdelete)
 * [`chectl server:logs`](#chectl-serverlogs)
 * [`chectl server:start`](#chectl-serverstart)
@@ -167,6 +168,27 @@ OPTIONS
 ```
 
 _See code: [@oclif/plugin-help](https://github.com/oclif/plugin-help/blob/v2.2.1/src/commands/help.ts)_
+
+## `chectl server:debug`
+
+Enable local debug of Eclipse Che server
+
+```
+USAGE
+  $ chectl server:debug
+
+OPTIONS
+  -h, --help                               show CLI help
+
+  -n, --chenamespace=chenamespace          [default: che] Kubernetes namespace where Che server is supposed to be
+                                           deployed
+
+  --debug-port=debug-port                  [default: 8000] Eclipse Che Server debug port
+
+  --listr-renderer=default|silent|verbose  [default: default] Listr renderer
+```
+
+_See code: [src/commands/server/debug.ts](https://github.com/che-incubator/chectl/blob/v0.0.2/src/commands/server/debug.ts)_
 
 ## `chectl server:delete`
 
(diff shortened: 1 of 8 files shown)


src/api/kube.ts
```diff
@@ -956,6 +958,7 @@ export class KubeHelper {
       const imageAndTag = cheImage.split(':', 2)
       yamlCr.spec.server.cheImage = imageAndTag[0]
       yamlCr.spec.server.cheImageTag = imageAndTag.length === 2 ? imageAndTag[1] : 'latest'
+      yamlCr.spec.server.cheDebug = flags.debug !== undefined ? flags.debug.toString() : 'false'
```
AndrienkoAleksandr: It would be nice to apply type boolean check.

src/tasks/che.ts
```diff
@@ -554,4 +555,42 @@ export class CheTasks {
...
+            throw new Error(`Che Server pod not found in the namespace '${flags.chenamespace}'`)
+          }
+          ctx.podName = chePods.items[0].metadata!.name!
+          task.title = `${task.title}...done`
+        }
+      },
+      {
+        title: 'Check if debug mode is enabled',
+        task: async (ctx: any, task: any) => {
+          const res = await execa(`kubectl exec ${ctx.podName} -n ${flags.chenamespace} -- printenv CHE_DEBUG_SERVER `, { timeout: 10000, reject: false, shell: true })
```
AndrienkoAleksandr: I guess we can get this information from config map.

AndrienkoAleksandr: Or you can get env variable from pods container: https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#pods-inject-envars-yaml . Container resource object has field env

README.md
```diff
@@ -263,6 +285,9 @@ OPTIONS
       [default: quay.io/eclipse/che-operator:nightly] Container image of the operator. This parameter is used only when 
       the installer is the operator
 
+  --debug
+      Enables the debug mode for Che server
```
sleshchenko: It could be useful to put here some info like: 

> after that, you'll be able to use `server:debug` command to connect to your Che Server via debug protocol
> But note that it's not recommended to do it on publicly available instances since there are a few modifications in Che Server configuration which does your Che non-production-ready.

I'm not sure about wording and messages. Just review it and think whether it really makes sense )

src/tasks/che.ts
```diff
@@ -554,4 +555,42 @@ export class CheTasks {
...
     ]
   }
+
+  debugTask(flags: any): ReadonlyArray<Listr.ListrTask> {
+    return [
+      {
+        title: 'Find Che Server pod',
+        task: async (ctx: any, task: any) => {
+          const chePods = await this.kube.listNamespacedPod(flags.chenamespace, undefined, this.cheSelector)
+          if (chePods.items.length === 0) {
```
sleshchenko: Does it make sense to check if there are more than 1 item?
It's possible after rescaling when one pod is terminating and another is starting...
picking the first up could be not a straightforward thing to do.

src/tasks/che.ts
```diff
@@ -554,4 +555,42 @@ export class CheTasks {
...
+      {
+        title: 'Check if debug mode is enabled',
+        task: async (ctx: any, task: any) => {
+          const res = await execa(`kubectl exec ${ctx.podName} -n ${flags.chenamespace} -- printenv CHE_DEBUG_SERVER `, { timeout: 10000, reject: false, shell: true })
+          if (res.exitCode !== 0) {
+            throw new Error(`'kubectl exec' failed with code: ${res.exitCode}, message: ${res.stderr}`)
+          }
+
+          if (res.stdout.toUpperCase() !== 'TRUE') {
+            throw new Error('Che Serve should be redeployed with \'--debug\' flag')
```
sleshchenko: Serve**r**

src/tasks/che.ts
tolusha: > Container resource object has field env

yes, but of env variables are bind from config map they aren't available in `env` field unfortunately.

src/tasks/che.ts
tolusha: Yeap, but more than 1 server is possible case.

PR #482, opened 2020-02-03
fix(helm): don't set workspace namespace for helm installer

sparkoo:
Signed-off-by: Michal Vala <mvala@redhat.com>

### What does this PR do?
Don't set namespace for workspaces when using helm installer. Helm charts already has reasonable default, which should be used and chectl should not interfere with it. Setting workspaces namespace can be implemented as new chectl parameter, but it's not in scope of this task.

https://github.com/eclipse/che/issues/15440

### DIFF
--- a/src/tasks/installers/helm.ts
+++ b/src/tasks/installers/helm.ts
@@ -331,7 +331,6 @@ error: E_COMMAND_FAILED`)
 
     setOptions.push(`--set global.ingressDomain=${flags.domain}`)
     setOptions.push(`--set cheImage=${flags.cheimage}`)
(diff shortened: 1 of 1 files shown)


tolusha:
@sparkoo 
Have you tried to deploy Che using `chectl` based on this PR?
I am wondering which namespace will be used in this case.

sparkoo:
@tolusha yes. It uses value from helm charts https://github.com/eclipse/che/blob/master/deploy/kubernetes/helm/che/values.yaml#L67
`<username>-che` which is what we want https://github.com/eclipse/che/issues/14614

PR #500, opened 2020-02-10
feat: Add a chectl preflight check step that verifies that the platform version is supported

tolusha:
### What does this PR do?
Add minimal versions check for 
- openshift
- kubernetes
- helm

https://github.com/eclipse/che/issues/15680

### DIFF
--- a/src/api/version.ts
+++ b/src/api/version.ts
@@ -79,15 +79,15 @@ export namespace VersionHelper {
     return getVersionWithKubectl('Server Version: ')
   }
 
-  export function checkMinimalK8sVersions(actualVersion: string): boolean {
+  export function checkMinimalK8sVersion(actualVersion: string): boolean {
     return checkMinimalVersion(actualVersion, MINIMAL_K8S_VERSION)
   }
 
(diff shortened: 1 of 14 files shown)


src/api/version.ts
```diff
@@ -0,0 +1,134 @@
...
+    const minimalMinor = parseInt(vers[1], 10)
+
+    return (actualMajor >= minimalMajor || (actualMajor === minimalMajor && actualMinor >= minimalMinor))
+  }
+
+  static getError(actualVersion: string, minimalVersion: string, component: string): Error {
+    return new Error(`The minimal supported version of ${component} is '${minimalVersion} but found '${actualVersion}'. To bypass version check use '--skip-version-check' flag.`)
+  }
+
+  private static async getVersionWithOC(versionPrefix: string): Promise<string | undefined> {
```
mmorhun: I think it's not `versionPrefix` (look at [removePrefix](https://github.com/che-incubator/chectl/pull/500/files#diff-a1b5de5b01da816530356f13e924fe22R131)), it is more like `componentLabel` or so.

PR #507, opened 2020-02-13
feat: add snippet about how to obtain user access...

nickboldt:
add snippet about how to obtain user access token when stopping a che/crw server

Change-Id: I4bdddabb6f2329264887e9380202e1b9e7118833

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,16 @@ OPTIONS
   --listr-renderer=default|silent|verbose  [default: default] Listr renderer
 ```
 
+To obtain a user access token from the indentity provider (eg, Keycloak or Red Hat Single Sign-On), using `oc` or `kubectl` with `curl` and `jq`:
+
+```
+KEYCLOAK_URL=$(kubectl get route/keycloak -n ${cheNamespace} -o jsonpath='{.spec.host}')
+USER_ACCESS_TOKEN=$(curl -X POST $KEYCLOAK_URL/auth/realms/che/protocol/openid-connect/token \
+  -H "Content-Type: application/x-www-form-urlencoded" \
+  -d "username=${username}" -d "password=${password}" 
+  -d "grant_type=password" -d "client_id=che-public" | jq -r .access_token)
+```
+
 _See code: [src/commands/server/stop.ts](https://github.com/che-incubator/chectl/blob/v0.0.2/src/commands/server/stop.ts)_
 
 ## `chectl server:update`


README.md
```diff
@@ -348,6 +348,16 @@ OPTIONS
   --listr-renderer=default|silent|verbose  [default: default] Listr renderer
 ```
 
+To obtain a user access token from the indentity provider (eg, Keycloak or Red Hat Single Sign-On), using `oc` or `kubectl` with `curl` and `jq`:
+
+```
+KEYCLOAK_URL=$(kubectl get route/keycloak -n ${cheNamespace} -o jsonpath='{.spec.host}')
+USER_ACCESS_TOKEN=$(curl -X POST $KEYCLOAK_URL/auth/realms/che/protocol/openid-connect/token \
+  -H "Content-Type: application/x-www-form-urlencoded" \
+  -d "username=${username}" -d "password=${password}"
```
dmytro-ndp: we need ending "\" in at the end of this line as well

README.md
```diff
@@ -348,6 +348,16 @@ OPTIONS
   --listr-renderer=default|silent|verbose  [default: default] Listr renderer
 ```
 
+To obtain a user access token from the indentity provider (eg, Keycloak or Red Hat Single Sign-On), using `oc` or `kubectl` with `curl` and `jq`:
+
+```
+KEYCLOAK_URL=$(kubectl get route/keycloak -n ${cheNamespace} -o jsonpath='{.spec.host}')
+USER_ACCESS_TOKEN=$(curl -X POST $KEYCLOAK_URL/auth/realms/che/protocol/openid-connect/token \
```
dmytro-ndp: Wrong url is here.
It should be `http://$KEYCLOAK_URL/auth/realms/che/protocol/openid-connect/token`

tolusha:
`README.md` is autogenerated by `yarn pack` command.
All changes will be vanished if they aren't added at the end of the doc.

nickboldt:
> README.md is autogenerated by yarn pack command.

So where do I add this information ? in the source code? in a separate README? In the Che docs so it's harder to find / not available from `--help` ?

tolusha:
In source code, in case of access token the description should be updated
https://github.com/che-incubator/chectl/blob/ab/patchcr/src/common-flags.ts#L32

and then update README.md with `yarn pack`

README.md
bscholtes: what about https and make curl trust self-signed certificates.

README.md
```diff
@@ -358,6 +358,16 @@ OPTIONS
   --listr-renderer=default|silent|verbose  [default: default] Listr renderer
 ```
 
+To obtain a user access token from the indentity provider (eg, Keycloak or Red Hat Single Sign-On), using `oc` or `kubectl` with `curl` and `jq`:
+
+```
+KEYCLOAK_URL=$(kubectl get route/keycloak -n ${cheNamespace} -o jsonpath='{.spec.host}')
```
AndrienkoAleksandr: I think we should notify, that it will work on Linux. For windows I don't think such shell obstructionist will work.

nickboldt: by "notify" you mean something other than a console error message?

README.md
nickboldt: Yeah, we should default to https. What's the curl URL for self-signed cert?

bscholtes: I mean is there a way to use --insecure to avoid issues like
  "Unable to communicate securely with peer: 
  requested domain name does not match the server's certificate"
in case of a poorly cert/PKI setup

PR #508, opened 2020-02-13
feat: Allows to override default values

tolusha:

### What does this PR do?
Allows to override default values of custom resources by providing a patch file.

patch.yaml
```yaml
spec:
  storage:
    pvcStrategy: 'per-workspace'
```

```bash
chectl server:start  --che-operator-cr-patch-yaml=patch.yaml ....
```

https://github.com/eclipse/che/issues/15684

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,8 +280,8 @@ OPTIONS
       [default: templates] Path to the templates folder
 
   --che-operator-cr-patch-yaml=che-operator-cr-patch-yaml
-      Path to a yaml file that overrides the default value a CheCluster used by the operator. This parameter is used only 
-      when the installer is the operator.
+      Path to a yaml file that overrides the default values in CheCluster CR used by the operator. This parameter is used 
+      only when the installer is the operator.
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -279,6 +279,10 @@ OPTIONS
   -t, --templates=templates
       [default: templates] Path to the templates folder
 
+  --che-operator-cr-patch=che-operator-cr-patch
```
sleshchenko: well, it's actually checluster-cr... Maybe it makes sense to name it like this then?

sleshchenko:
```suggestion
  --che-operator-cr-patch=che-operator-cr-patch.yaml
```

README.md
```diff
@@ -279,6 +279,10 @@ OPTIONS
...
+      Path to a yaml file that overrides the default value a CheCluster used by the operator. This parameter is used only
```
sleshchenko:
```suggestion
      Path to a yaml file that overrides the default values in CheCluster CR used by the operator. This parameter is used only 
```

PR #517, opened 2020-02-17
feat: Implement kube config injection on openshift cluster

mmorhun:

### What does this PR do?
Implements injection of Kubernetes config from host into Che workspace for Openshift infrastructure.
Also implements config injection for Openshift cluster with kubectl (yes, `kubectl` and `oc` is not interchangeable in all cases).

### How to test
Start your cluster, deploy Che on it. When ready run the following command:
```sh
chectl workspace:inject -k -c <container-name>
```

If one wants to test specific CLI tool (`oc` or `kubectl`) with the cluster, then one need to remove other options from CLI list in `getClusterClientCommand` function in `src/util.ts` file.

To be able to test in some local clusters which has 'bad' TLS certificate, one need to relogin with `--insecure-skip-tls-verify` option and run `chectl` (as described above) before autologout. For example, for CRC, the command might look like:
```sh
oc logout --insecure-skip-tls-verify && oc login -u kubeadmin -p BdTY8-Vra64-gD3cI-dAPS6 https://api.crc.testing:6443 --insecure-skip-tls-verify && ./run workspace:inject -k -c nodejs
```

https://github.com/eclipse/che/issues/15987

### DIFF
--- a/src/commands/workspace/inject.ts
+++ b/src/commands/workspace/inject.ts
@@ -13,6 +13,7 @@ import { Context } from '@kubernetes/client-node/dist/config_types'
 import { Command, flags } from '@oclif/command'
 import { string } from '@oclif/parser/lib/flags'
 import * as execa from 'execa'
+import * as fs from 'fs'
 import * as Listr from 'listr'
 import * as os from 'os'
 import * as path from 'path'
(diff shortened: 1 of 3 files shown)


src/commands/workspace/inject.ts
```diff
@@ -48,11 +50,21 @@ export default class Inject extends Command {
...
+
   async run() {
     const { flags } = this.parse(Inject)
     const notifier = require('node-notifier')
     const cheTasks = new CheTasks(flags)
 
+    // Initialize cluster CLI tool name
+    const command = getClusterClientCommand()
+    if (!command) {
+      throw new Error('No cluster CLI client is installed.')
```
tolusha: Is it better to throw the error inside `getClusterClientCommand` ?

AndrienkoAleksandr: +1

mmorhun: Done

PR #525, opened 2020-02-20
feat: Show warning for unused properties

tolusha:

### What does this PR do?
Show warning for unused properties

```bash
$ ./run server:start --platform=minikube  --installer=operator --che-operator-cr-yaml=cr.yaml --cheimage=test --debug
  ✔ Verify Kubernetes API...OK
  ✔ 👀  Looking for an already existing Eclipse Che instance
    ✔ Verify if Eclipse Che is deployed into namespace "che"...it is not
 ›   Warning: --che-operator-cr-yaml is used. The following flag(s) will be ignored: --cheimage	--debug
```

https://github.com/eclipse/che/issues/15072


src/commands/server/start.ts
```diff
@@ -186,6 +186,20 @@ export default class Start extends Command {
 
   checkPlatformCompatibility(flags: any) {
     // matrix checks
+    if (flags.installer === 'operator' && flags['che-operator-cr-yaml']) {
+      this.warn('--che-operator-cr-yaml is used. The following flags will be ignored:')
```
sleshchenko: Does it make sense not to show this message if user did not specify any of the listed parameters?

src/commands/server/start.ts
```diff
@@ -186,6 +186,23 @@ export default class Start extends Command {
...
+    if (flags.installer === 'operator' && flags['che-operator-cr-yaml']) {
+      let msg = ''
+      msg += flags['plugin-registry-url'] ? '\t--plugin-registry-url' : ''
+      msg += flags['devfile-registry-url'] ? '\t--devfile-registry-url' : ''
+      msg += flags['postgres-pvc-storage-class-name'] ? '\t--postgres-pvc-storage-class-name' : ''
+      msg += flags['workspace-pvc-storage-class-name'] ? '\t--workspace-pvc-storage-class-name' : ''
+      msg += flags['self-signed-cert'] ? '\t--self-signed-cert' : ''
+      msg += flags['os-oauth'] ? '\t--os-oauth' : ''
+      msg += flags.tls ? '\t--ls' : ''
+      msg += flags.cheimage ? '\t--cheimage' : ''
```
sleshchenko: Screenshot_20200224_092914

It's not critical but cheimage is printed even when user does not specify it.

src/commands/server/start.ts
```diff
@@ -186,6 +186,23 @@ export default class Start extends Command {
 
   checkPlatformCompatibility(flags: any) {
     // matrix checks
+    if (flags.installer === 'operator' && flags['che-operator-cr-yaml']) {
```
sleshchenko: Is it actual also when user use `--che-operator-cr-patch-yaml`?

src/commands/server/start.ts
sleshchenko: Screenshot_20200224_093135

PR #539, opened 2020-02-26
refactor: Move cert-manager deployment templates into Che repository

mmorhun:

### What does this PR do?
Moves templates related to Cert Manager deployment from CheCtl repository into Che repository.
Cert Manager installation yaml file is moved from templates folder (as it is not a template) into installers folder.

**Depends on:** https://github.com/eclipse/che/pull/16144


.gitignore
```diff
@@ -15,3 +15,4 @@ tsconfig.tsbuildinfo
 templates/kubernetes
 templates/che-operator
 templates/minishift-addon
+templates/cert-manager
```
benoitf: I think we don't need it no ?

package.json
```diff
@@ -65,6 +65,7 @@
   "files": [
     "/bin",
     "/lib",
+    "/installers",
```
benoitf: why do we need installers there ?

PR #543, opened 2020-02-27
fix: Improve deployment of Cert Manager to support older Kubernetes versions

mmorhun:

### What does this PR do?
Fixes deployment of Cert Manager to be able to use Kubernetes v1.15.4 or older.

Tested with Kubernetes v1.12.0


dmytro-ndp:
This PR has fixed an issue https://github.com/eclipse/che/issues/16158

PR #555, opened 2020-03-04
fix: reuse variables so we have fewer strings to replace in downstream

nickboldt:
reuse this.devfileRegistrySelector and this.pluginRegistrySelector so we have fewer strings to replace in downstream

Change-Id: Ib8b63a2b225f4364a300d2e57eb15a401e53f497


nickboldt:
e2e_tests

nickboldt:
pr_check

PR #556, opened 2020-03-04
fix: use consistent variable name for all minishiftAddonTasks

nickboldt:
use consistent variable name for all minishiftAddonTasks

Change-Id: I86ea46af0218f4332e58a8ecefc820209eb2cb02

### DIFF
--- a/src/commands/server/delete.ts
+++ b/src/commands/server/delete.ts
@@ -41,7 +41,7 @@ export default class Delete extends Command {
 
     const apiTasks = new ApiTasks()
     const helmTasks = new HelmTasks(flags)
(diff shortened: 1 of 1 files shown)


nickboldt:
Three files use minishiftAddonTasks:

src/tasks/installers/installer.ts
test/tasks/installers/minishift-addon.test.ts
and 
src/commands/server/delete.ts

But in the first to the var is `minishiftAddonTasks` whereas in delete, it's `msAddonTasks`.

Being consistent makes it easier to filter this functionality out when generating crwctl downstream code.

flacatus:
[test]

PR #562, opened 2020-03-06
fix: Fall back to CR default value of tls support if tls flag is omitted

mmorhun:

Backports https://github.com/che-incubator/chectl/pull/561 to `7.9.x` branch


davidfestal:
@tolusha yes, in the `che-operator` GH repo, as well as the CR in the `deploy/` folder in the same repo.

PR #568, opened 2020-03-09
fix:  use this.fooSelector instead of invalid selectors

nickboldt:
fix deployment selections to use this.fooSelector instead of invalid selectors

Change-Id: Ie97fc3c6378e158032109ee6e4ee5e23053146ab


nickboldt:
```
/bin/bash: .ci/cico_chectl_prcheck.sh: No such file or directory
```

Nice one, https://ci.centos.org/job/devtools-chectl-pr-check-chectl/82/console

tolusha:
@nickboldt 
you replaced
'keycloak' -> `app=keycloak` where deployment name is expected, will it work?

PR #573, opened 2020-03-11
feat: Support deploying in single user mode with operator

tolusha:

### What does this PR do?
Support deploying in single user mode with operator

https://github.com/eclipse/che/issues/13420
https://github.com/eclipse/che-operator/pull/185

### DIFF
--- a/src/commands/server/start.ts
+++ b/src/commands/server/start.ts
@@ -152,7 +152,7 @@ export default class Start extends Command {
     return path.join(__dirname, '../../../templates')
   }
 
-  static setPlaformDefaults(flags: any) {
+  setPlaformDefaults(flags: any) {
     if (flags.platform === 'minishift') {
       if (!flags.multiuser && flags.installer === '') {
         flags.installer = 'minishift-addon'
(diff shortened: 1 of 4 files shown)


src/commands/server/start.ts
```diff
@@ -205,6 +204,7 @@ export default class Start extends Command {
       flags.cheimage && ignoredFlags.push('--cheimage')
       flags.debug && ignoredFlags.push('--debug')
       flags.domain && ignoredFlags.push('--domain')
+      flags.multiuser && ignoredFlags.push('--multiuser')
```
mmorhun: Shouldn't chectl flags have higher priority over config file?

tolusha: They have a higher priority over defaults but a less priority in comparison with provided CR

src/tasks/installers/installer.ts
```diff
@@ -80,9 +80,8 @@ export class InstallerTasks {
     } else if (flags.installer === 'operator') {
       title = '🏃‍  Running the Eclipse Che operator'
       task = () => {
-        // The operator installs Eclipse Che in multiuser mode only
+        // The operator installs Eclipse Che in multiuser mode by default
```
AndrienkoAleksandr: Maybe, make sense to have a flag single user, to have ability fast deploy Che in this mode without playing with CR yaml or CR yaml patch?

tolusha: Single-user mode for operator is what I would like to hide.
It is just to deprecate `deploy.sh` script

PR #574, opened 2020-03-11
feat: Add workspace:list command

tolusha:

### What does this PR do?
Add `workspace:list` command
```
  ✔ Verify Kubernetes API...OK
  ✔ Verify if namespace che exists
  ✔ Retrieving Eclipse Che server URL... http://che-che.192.168.99.100.nip.io
  ✔ Eclipse Che status check
  ✔ Get workspaces... done
Id      Name    Namespace       Status  Last modified
workspaceadm0rrdlo2kt9hj9	golang-5ssqg	che	STARTING	Wed Mar 11 2020 12:40:25 GMT+0200 (Eastern European Standard Time)
```

https://github.com/eclipse/che/issues/15071

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,6 +456,10 @@ OPTIONS
   --access-token=access-token              Eclipse Che OIDC Access Token
 
   --listr-renderer=default|silent|verbose  [default: default] Listr renderer
+
+  --max-items=max-items                    [default: 30] The maximum number of workspaces to return
+
+  --skip-count=skip-count                  The number of workspaces to skip
 ```
 
 _See code: [src/commands/workspace/list.ts](https://github.com/che-incubator/chectl/blob/v0.0.2/src/commands/workspace/list.ts)_
--- a/README.md
+++ b/README.md
@@ -456,10 +456,6 @@ OPTIONS
   --access-token=access-token              Eclipse Che OIDC Access Token
 
   --listr-renderer=default|silent|verbose  [default: default] Listr renderer
-
-  --max-items=max-items                    [default: 30] The maximum number of workspaces to return
-
-  --skip-count=skip-count                  The number of workspaces to skip
 ```
 
 _See code: [src/commands/workspace/list.ts](https://github.com/che-incubator/chectl/blob/v0.0.2/src/commands/workspace/list.ts)_
--- a/src/commands/workspace/list.ts
+++ b/src/commands/workspace/list.ts
@@ -9,8 +9,13 @@
  **********************************************************************/
 
 import { Command, flags } from '@oclif/command'
+import { cli } from 'cli-ux'
+import * as Listrq from 'listr'
 
+import { CheHelper } from '../../api/che'
 import { accessToken, cheNamespace, listrRenderer } from '../../common-flags'
+import { CheTasks } from '../../tasks/che'
+import { ApiTasks } from '../../tasks/platforms/api'
 
 export default class List extends Command {
   static description = 'list workspaces'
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -456,6 +456,10 @@ OPTIONS
   --access-token=access-token              Eclipse Che OIDC Access Token
 
   --listr-renderer=default|silent|verbose  [default: default] Listr renderer
+
+  --max-items=max-items                    [default: 30] The maximum number of workspaces to return
+
+  --skip-count=skip-count                  The number of workspaces to skip
```
sleshchenko: 1. I don't time they're parameters for CLI... Could we check like how kubectl, oc does and do the same... Maybe it makes sense to fetch all workspaces page by page internally and show them?

2. If you think we need paging here, won't it be more convenient to use --page -p, --page-size?
like `workspace:list -p=3` instead `workspace:list --skip-count=90`, I know it's not an issue to calculate 30*3 in mind but it makes sense to let people learn math with other tools but not chectl.

I don't think users will make many workspaces, so my personal :+1: for item 1

tolusha: Make sense!

src/commands/workspace/list.ts
```diff
@@ -19,30 +23,53 @@ export default class List extends Command {
...
+    try {
+      await tasks.run(ctx)
+      this.printWorkspaces(ctx.workspaces)
+    } catch (error) {
+      this.error(error)
+    }
+  }
+
+  private printWorkspaces(workspaces: any): void {
+    this.log('Id\tName\tNamespace\tStatus\tLast modified')
```
mmorhun: I think we may hardcode right number of tabs here. Then we'll have table-like header.

src/commands/workspace/list.ts
```diff
@@ -19,30 +23,53 @@ export default class List extends Command {
...
+    for (const workspace of workspaces) {
+      const lastModified = new Date(parseInt(workspace.attributes.updated, 10))
```
mmorhun: Does it make sense to add `Creation date`?

mmorhun: Do we need expanded time zone text? Isn't GMT notation is enough?

src/commands/workspace/list.ts
tolusha: It depends on length of data

src/commands/workspace/list.ts
tolusha: make sense

src/commands/workspace/list.ts
mmorhun: Well, we may predict that workspace id is always as long as `workspaceadm0rrdlo2kt9hj9` and the same for other fields except workspace name. Here we can either truncate it or move to the last column.

tolusha: You never know the namespace length.
In general case it is needed to implement some complex logic.
I would keep as is for know

tolusha:
[test]

tolusha:
<pre>[10:52:22] tolusha@fractal ~/projects/chectl/bin (ab/wlist) $ ./run workspace:list
  <font color="#4D9A05">✔</font> Verify Kubernetes API...OK
  <font color="#4D9A05">✔</font> Verify if namespace che exists
  <font color="#4D9A05">✔</font> Retrieving Eclipse Che server URL... https://che-che.192.168.99.100.nip.io
  <font color="#4D9A05">✔</font> Eclipse Che status check
  <font color="#4D9A05">✔</font> Get workspaces... done
<b>Id                        Name              Namespace Status  Created                  Updated                  </b>
workspacefqq2289aq4fjdm2s golang-o7db4      che-che   STOPPED 2020-03-17T07:54:40.293Z 2020-03-17T07:54:41.372Z 
workspaceiof4jt5u8f2vec7v java-gradle-o8fwf che-che   STOPPED 2020-03-17T07:54:54.239Z </pre>

PR #592, opened 2020-03-19
feat: Add workspace:create command instead of workspace:start

mmorhun:
### What does this PR do?
Adds new `workspace:create` command to `chectl`, which is supposed to replace `worspace:start` which become deprecated.

`workspace:create` command has `--start` option which starts the workspace right after creation.

In case of using `workspace:start` command, deprecation warning will be shown:
output

https://github.com/eclipse/che/issues/14344

### DIFF
--- a/src/api/che.ts
+++ b/src/api/che.ts
@@ -205,14 +205,14 @@ export class CheHelper {
     let response
 
     const headers: {[key: string]: string} = {}
-    if (accessToken && accessToken.length > 0) {
+    if (accessToken) {
       headers.Authorization = accessToken
     }
     try {
       response = await this.axios.post(endpoint, undefined, { headers })
     } catch (error) {
       if (error.response && error.response.status === 404) {
-        throw new Error('E_WORKSPACE_NOT_EXIST')
+        throw new Error(`E_WORKSPACE_NOT_EXIST - workspace with "${workspaceId}" id doesn't exist`)
       } else {
         throw this.getCheApiError(error, endpoint)
       }
--- a/src/api/che.ts
+++ b/src/api/che.ts
@@ -199,12 +199,17 @@ export class CheHelper {
     }
   }
 
-  async startWorkspace(cheNamespace: string, workspaceId: string): Promise<void> {
+  async startWorkspace(cheNamespace: string, workspaceId: string, accessToken?: string): Promise<void> {
     const cheUrl = await this.cheURL(cheNamespace)
     const endpoint = `${cheUrl}/api/workspace/${workspaceId}/runtime`
     let response
+
+    const headers: {[key: string]: string} = {}
+    if (accessToken && accessToken.length > 0) {
+      headers.Authorization = accessToken
+    }
     try {
-      response = await this.axios.post(endpoint)
+      response = await this.axios.post(endpoint, undefined, { headers })
     } catch (error) {
       if (error.response && error.response.status === 404) {
         throw new Error('E_WORKSPACE_NOT_EXIST')
--- a/src/api/che.ts
+++ b/src/api/che.ts
@@ -1,5 +1,5 @@
 /*********************************************************************
- * Copyright (c) 2019 Red Hat, Inc.
+ * Copyright (c) 2019-2020 Red Hat, Inc.
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -7,6 +7,8 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  **********************************************************************/
+
+import { che as chetypes } from '@eclipse-che/api'
 import { CoreV1Api, KubeConfig, V1Pod, Watch } from '@kubernetes/client-node'
 import axios, { AxiosInstance } from 'axios'
 import * as cp from 'child_process'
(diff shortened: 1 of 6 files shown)


tolusha:
1. Remove flag from both commands:
`'listr-renderer': listrRenderer`
and use 
`{ renderer: 'silent' }`
2. Use `cli.warn`  or `cli.log` to print information

tolusha:
3. `yarn pack` to update `README.md`

yarn.lock
```diff
@@ -123,6 +123,10 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
+"@eclipse-che/api@latest":
+  version "7.5.0-SNAPSHOT"
```
tolusha: looks strange

src/commands/workspace/start.ts
```diff
@@ -8,72 +8,29 @@
...
-    'listr-renderer': listrRenderer
-  }
-
-  async checkToken(flags: any, ctx: any) {
-    if (ctx.isAuthEnabled && !flags['access-token']) {
-      this.error('E_AUTH_REQUIRED - Eclipse Che authentication is enabled and an access token need to be provided (flag --access-token).')
-    }
-  }
+export default class Start extends Create {
+  static description = 'Creates and starts workspace from devfile'
```
tolusha: I believe that this command is still used.
So, let's keep the current functionality + add warning...

src/commands/workspace/create.ts
```diff
@@ -0,0 +1,110 @@
...
+    help: flags.help({ char: 'h' }),
+    chenamespace: cheNamespace,
+    devfile: string({
+      char: 'f',
+      description: 'path or URL to a valid devfile',
+      env: 'DEVFILE_PATH',
+      required: true,
+    }),
+    name: string({
+      description: 'workspace name: overrides the workspace name to use instead of the one defined in the devfile. Works only for devfile',
```
tolusha: The devfile is the only way. The description should be adjusted

src/commands/workspace/create.ts
```diff
@@ -0,0 +1,110 @@
...
+  static description = 'Creates a workspace from devfile'
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    chenamespace: cheNamespace,
+    devfile: string({
+      char: 'f',
+      description: 'path or URL to a valid devfile',
+      env: 'DEVFILE_PATH',
+      required: true,
```
tolusha: We can make it optional. If flag isn't set then let's use the devfile from the current directory

src/commands/workspace/create.ts
```diff
@@ -0,0 +1,110 @@
...
+import { Command, flags } from '@oclif/command'
+import { boolean, string } from '@oclif/parser/lib/flags'
+import { cli } from 'cli-ux'
+import * as Listr from 'listr'
+import * as notifier from 'node-notifier'
+
+import { CheHelper } from '../../api/che'
+import { accessToken, cheNamespace, listrRenderer } from '../../common-flags'
+export default class Create extends Command {
+  static description = 'Creates a workspace from devfile'
```
tolusha: `from a devfile`

yarn.lock
AndrienkoAleksandr: Hmm. It is really latest version... https://www.npmjs.com/package/@eclipse-che/api

mmorhun: Nothing strange. I just used `latest` in package json, so it was resolved to `7.5.0-SNAPSHOT`

tolusha: Why do we need this at all?

AndrienkoAleksandr: I think @mmorhun wants to use dts types for types

mmorhun: Yes, I think it is good to have response structure typings generated from Che servers DTOs

tolusha: But is it outdated version?

AndrienkoAleksandr: @azatsarynnyy @vparfonov  Do we have newer version of the @eclipse-che/api or release mechanizm become broken after moving to the centos ci?

azatsarynnyy: @AndrienkoAleksandr do you know where's the repo with the sources of `@eclipse-che/api`?

AndrienkoAleksandr: I think publish here https://github.com/eclipse/che/tree/master/typescript-dto

azatsarynnyy: Sorry @AndrienkoAleksandr but I have no idea who is responsible now for `@eclipse-che/api`. Try to ask in Eclipse Che MM channel tomorrow.

src/commands/workspace/create.ts
mmorhun: Are you suggesting to remove this option at all?

tolusha: Just remove that
`Works only for devfile`

yarn.lock
mmorhun: I think it should a part of release process.

tolusha: Let's not use outdated version.

src/commands/workspace/start.ts
mmorhun: That's exactly what I do.

yarn.lock
mmorhun: I don't think there were some major changes. But we need to raise this question.

src/commands/workspace/create.ts
mmorhun: ok

yarn.lock
mmorhun: I've create issue for this problem: https://github.com/eclipse/che/issues/16436

src/api/che.ts
```diff
@@ -197,13 +199,37 @@ export class CheHelper {
...
   }
 
-  async createWorkspaceFromDevfile(namespace: string | undefined, devfilePath = '', workspaceName: string | undefined, accessToken = ''): Promise<string> {
+  async startWorkspace(cheNamespace: string, workspaceId: string, accessToken?: string): Promise<void> {
+    const cheUrl = await this.cheURL(cheNamespace)
+    const endpoint = `${cheUrl}/api/workspace/${workspaceId}/runtime`
+    let response
+
+    const headers: {[key: string]: string} = {}
+    if (accessToken && accessToken.length > 0) {
```
tolusha: `if (accessToken) ` enough?

src/api/che.ts
```diff
@@ -197,13 +199,37 @@ export class CheHelper {
...
+      headers.Authorization = accessToken
+    }
+    try {
+      response = await this.axios.post(endpoint, undefined, { headers })
+    } catch (error) {
+      if (error.response && error.response.status === 404) {
```
tolusha: `getCheApiError` function actually catches  404 error (rebase on master pls)

src/commands/workspace/create.ts
```diff
@@ -0,0 +1,122 @@
...
+          if (ctx.isAuthEnabled && !flags['access-token']) {
+            this.error('E_AUTH_REQUIRED - Eclipse Che authentication is enabled and an access token needs to be provided (flag --access-token).')
+          }
+
+          let devfilePath: string
+          if (flags.devfile) {
+            devfilePath = flags.devfile
+          } else if (fs.existsSync('devfile.yaml')) {
+            devfilePath = 'devfile.yaml'
+          } else if (fs.existsSync('devfile.yml')) {
```
tolusha: duplication ?

mmorhun: No, the extension is different

src/api/che.ts
mmorhun: yes

src/api/che.ts
mmorhun: Not only...

mmorhun: I want to handle 404 error in another way for this specific request.

mmorhun:
Rebased

PR #598, opened 2020-03-23
fix: Fix starting multiple che instances in the same cluster

AndrienkoAleksandr:
### What does this PR do?
Provide unique clusterrole and clusterrole binding per che instance in the same cluster. Use namespace suffix for this purpose: `${che-role-name}-namespace` and `${che-rolebinding-name}-namespace`

https://github.com/eclipse/che/issues/15345
Mostly for https://issues.redhat.com/browse/CRW-505?focusedCommentId=13816159&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel

### DIFF
--- a/src/tasks/installers/operator.ts
+++ b/src/tasks/installers/operator.ts
@@ -35,6 +35,8 @@ export class OperatorTasks {
    * Returns tasks list which perform preflight platform checks.
    */
   startTasks(flags: any, command: Command): Listr {
+    const clusterRoleName = `${this.operatorClusterRole}-${flags.chenamespace}`
+    const clusterRoleBindingName = `${this.operatorClusterRoleBinding}-${flags.chenamespace}`
     const che = new CheHelper(flags)
     const kube = new KubeHelper(flags)
     return new Listr([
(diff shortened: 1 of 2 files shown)


AndrienkoAleksandr:
I still want to try better handle code part with clean up CRD. Currently I removed clean up CRD on remove Che, because CRD has global scope. We can not create CRD in namespace scope https://github.com/kubernetes/kubernetes/issues/65551 per Che installation. When we remove CRD to clean one Che installation we actually remove CRD for all  Che instances and breaks all installation in the same cluster...

src/tasks/installers/operator.ts
```diff
@@ -376,15 +382,6 @@ export class OperatorTasks {
         }
       }
     },
-    {
```
davidfestal: I understand that we should not delete the CRD when it might still be used by other installations.
OTOH for now, on the `chectl server:start`, the CRD is only created if it didn't exist as shown [here](https://github.com/che-incubator/chectl/blob/e1af325ea7ab114a7a6d04c423b3cf2a2a99cf93/src/tasks/installers/operator.ts#L134).
That means that deleting a Che server with a `chectl` version and reinstalling with a new version of `chectl` will not update the CRD anymore.
That's a hard-to-solve question here, since we can manage only one version of the CRD.
That's a typical case where OLM-based installation has a real value, because the logic of CRD update is managed by OLM itself.

davidfestal: cc @l0rd

tolusha: @davidfestal 
If we are not going to delete CRD file, I would suggest to update it every time we install che using checlt

tolusha: It allows to keep CRD up to date

PR #599, opened 2020-03-24
feat: Add basic TLS configuration checks

mmorhun:

### What does this PR do?
Adds basic checks for TLS configuration of operator installer:
 - For Kubernetes infrastructure family in case of TLS it is required to have pre-created secret. If it is not provided the expanded error message will be shown along with link to the corresponding docs section:
tls-checks

 - Does basic check for `--self-signed-flag` option presence and automatically adds the flag if it is clear that self-signed certificate is used. However this check doesn't perform certificate analyzation directly, so it cannot catch all invalid cases, but only most frequent ones. I think that full certificate analyzation is out of scope of this PR, but left todo mark in the code.

Also corrects description for `--tls` flag.

https://github.com/eclipse/che/issues/16396

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ OPTIONS
                            Note, this option is turned on by default.
                            For Kubernetes infrastructure, it is required to provide own certificate: 'che-tls' secret with 
       TLS certificate must be pre-created in the configured namespace.
+                           The only exception is Helm installer. In that case the secret will be generated automatically.
                            For OpenShift, router will use default cluster certificates.
                            If the certificate is self-signed, '--self-signed-cert' option should be provided, otherwise 
       Che won't be able to start.
--- a/README.md
+++ b/README.md
@@ -273,10 +273,14 @@ OPTIONS
 
   -s, --tls
       Enable TLS encryption.
-                           Note, that this option is turned on by default for kubernetes infrastructure.
-                           If it is needed to provide own certificate, 'che-tls' secret with TLS certificate must be 
-      created in the configured namespace. Otherwise, it will be automatically generated.
+                           Note, this option is turned on by default.
+                           For Kubernetes infrastructure, it is required to provide own certificate: 'che-tls' secret with 
+      TLS certificate must be pre-created in the configured namespace.
                            For OpenShift, router will use default cluster certificates.
+                           If the certificate is self-signed, '--self-signed-cert' option should be provided, otherwise 
+      Che won't be able to start.
+                           Please see docs for more details: 
+      https://www.eclipse.org/che/docs/che-7/setup-che-in-tls-mode-with-self-signed-certificate/
 
   -t, --templates=templates
       [default: templates] Path to the templates folder
--- a/src/tasks/installers/operator.ts
+++ b/src/tasks/installers/operator.ts
@@ -68,7 +68,7 @@ export class OperatorTasks {
         task: async (_: any, task: any) => {
           // Che is being deployed on Kubernetes infrastructure
 
-          if (!(flags.tls || await this.checkCrForTls(flags))) {
+          if (! await this.checkTlsMode(flags)) {
             // No TLS mode, skip this check
             return
           }
@@ -94,7 +94,7 @@ export class OperatorTasks {
         // If the flag is set no need to check if it is required
         skip: () => flags['self-signed-cert'],
         task: async (_: any, task: any) => {
-          if (!(flags.tls || await this.checkCrForTls(flags))) {
+          if (! await this.checkTlsMode(flags)) {
             // No TLS mode, skip this check
             return
           }
@@ -119,6 +119,8 @@ export class OperatorTasks {
           }
 
           // TODO check the secret certificate if it is commonly trusted.
+          cli.info('TLS mode is turned on, however we failed to determine whether self-signed certificate is used. \n\
+                   Please rerun chectl with "--self-signed-cert" option if it is the case, otherwise Eclipse Che will fail to start.')
         }
       },
       {
@@ -563,7 +565,7 @@ export class OperatorTasks {
    * Checks if TLS is disabled via operator custom resource.
    * Returns true if TLS is enabled (or omitted) and false if it is explicitly disabled.
    */
-  private async checkCrForTls(flags: any): Promise<boolean> {
+  private async checkTlsMode(flags: any): Promise<boolean> {
     if (flags['che-operator-cr-yaml']) {
       const cheOperatorCrYamlPath = flags['che-operator-cr-yaml']
       if (fs.existsSync(cheOperatorCrYamlPath)) {
@@ -582,6 +584,11 @@ export class OperatorTasks {
           return false
         }
       }
+    }
+
+    // If tls flag is undefined we suppose that tls is turned on
+    if (flags.tls === false) {
+      return false
     }
 
     // TLS is on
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -273,10 +273,14 @@ OPTIONS
 
   -s, --tls
       Enable TLS encryption.
-                           Note, that this option is turned on by default for kubernetes infrastructure.
-                           If it is needed to provide own certificate, 'che-tls' secret with TLS certificate must be 
-      created in the configured namespace. Otherwise, it will be automatically generated.
+                           Note, this option is turned on by default.
+                           For Kubernetes infrastructure, it is required to provide own certificate: 'che-tls' secret with
```
sleshchenko: does not kubectl generate it if it's missing?

src/tasks/installers/operator.ts
```diff
@@ -60,6 +61,66 @@ export class OperatorTasks {
...
+        task: async (_: any, task: any) => {
+          if (!(flags.tls || await this.checkCrForTls(flags))) {
+            // No TLS mode, skip this check
+            return
+          }
+
+          const warningMessage = 'Self-signed certificate is used, so "--self-signed-cert" option is required. Added automatically.'
+
+          const platform = flags.platform
+          if (platform === 'minikube' || platform === 'crc' || platform === 'minishift') {
```
sleshchenko: QE and from time to time I use `https://console-openshift-console.apps.ocp43.codereadyqe.com/` where self-singed certs are used. And AFAIU it's not covered in this PR, so as before I will get broken Che if run `chectl server:start --installer=operator --platform=openshift` :disappointed: 
If it's not easy to determinate - let's ask users if cluster uses self-signed certs or not, like
```
TLS is used by default, does https://console-openshift-console.apps.ocp43.codereadyqe.com/ use self-signed certificate? Y/n
...
```
```

File: README.md (hunk 1 above)
Comment (mmorhun) [reply README.md]: I don't understand your question completely, could you clarify please?
If you are talking about helm installer - then yes, it will generate the secret and set self-signed flag, but these checks are added to operator tasks list.

File: src/tasks/installers/operator.ts (hunk 2 above)
Comment (mmorhun) [reply src/tasks/installers/operator.ts]: Thank you for deep review! Yes, you are right, however I had mentioned that as advanced case in the PR description, maybe not clear enough, sorry.
I may add this prompt, but I am not sure it is right way to go. It would bring another problems: how it would work in case of automated installation, say on PR check? new flag to disable this check? I think we have to implement certificate verification, which is separate issue. Until that users have to read docs as it was decided on planning...

**Comment by mmorhun** [comment]
[test]

File: src/tasks/installers/operator.ts (hunk 2 above)
Comment (sleshchenko) [reply src/tasks/installers/operator.ts]: >  Yes, you are right, however I had mentioned that as advanced case in the PR description, maybe not clear enough, sorry.

I saw it but still I imagine my emotions again and again when I remember your PR after figuring out that my Che is broken and I must reinstall it :cry: 

> It would bring another problems: how it would work in case of automated installation, say on PR check? new flag to disable this check? 

Ask that only if we enable tls by default
If user specifies `--tls` then we can blame themselves in missing `--self-signed-cert` :smiling_imp:

File: src/tasks/installers/operator.ts [hunk 3]
```diff
@@ -60,6 +61,66 @@ export class OperatorTasks {
           }
         }
       },
+      {
+        title: 'Checking for pre-created TLS secret',
+        // In case of Openshift infrastructure the certificate from cluster router will be used, so no need in the `che-tls` secret.
+        skip: () => !isKubernetesPlatformFamily(flags.platform),
```
tolusha: Is it better `isOpenshiftplatform` ?

README.md
sleshchenko: But it's help output of server:start command, right?
Not helm or operator installer specified by user at this stage...
As an user I would read it and expect that secret must be created on K8s before I run `chectl server:start --installer=helm --platform=kubernetes`. Sorry if I missed something or is still unclear.

src/tasks/installers/operator.ts
```diff
@@ -60,6 +61,66 @@ export class OperatorTasks {
...
+        task: async (_: any, task: any) => {
+          // Che is being deployed on Kubernetes infrastructure
+
+          if (!(flags.tls || await this.checkCrForTls(flags))) {
+            // No TLS mode, skip this check
```
tolusha: `flag.tls` is ignored if CR file is used.
so basically `flag.tls` make sense when `flags['che-operator-cr-yaml']` is not defined.

src/tasks/installers/operator.ts
mmorhun: To me it will sound like is `platform === 'openshift'`

src/tasks/installers/operator.ts
mmorhun: I'll add a warning in case when it is not clear if self-signed certificate used.

src/tasks/installers/operator.ts
tolusha: I mean `isOpenshiftPlatformFamily`

mmorhun: Yes, exactly. To me, if I rename `isOpenshiftPlatformFamily` to `isOpenshiftPlatform`, it would like the method just checks single value: `platform === 'openshift'`. So, I would prefer to keep the name as is.

src/tasks/installers/operator.ts
mmorhun: I see what you mean. Will change the priority.

src/tasks/installers/operator.ts
mmorhun: > Ask that only if we enable tls by default

So always

README.md
mmorhun: Agree, it is worth to add a note about Helm. Thank you for pointing out.

src/tasks/installers/operator.ts
mmorhun: I know that we have uncovered that case but it is not trivial and it would be better to handle it in a separate PR.

src/tasks/installers/operator.ts
```diff
@@ -60,6 +61,68 @@ export class OperatorTasks {
...
+          if (flags.domain && (flags.domain.endsWith('nip.io') || flags.domain.endsWith('xip.io'))) {
+            // It is not possible to use real certificate with *.nip.io and similar services
+            cli.warn(warningMessage)
+            flags['self-signed-cert'] = true
+            task.title = `${task.title}... self-signed`
+            return
+          }
+
+          // TODO check the secret certificate if it is commonly trusted.
+          cli.info('TLS mode is turned on, however we failed to determine whether self-signed certificate is used. \n\
```
tolusha: warn?

src/tasks/installers/operator.ts
```diff
@@ -497,4 +560,38 @@ export class OperatorTasks {
...
+      if (fs.existsSync(cheOperatorCrPatchYamlPath)) {
+        const crPatch = yaml.safeLoad(fs.readFileSync(cheOperatorCrPatchYamlPath).toString())
+        if (crPatch && crPatch.spec && crPatch.spec.server && crPatch.spec.server.tlsSupport === false) {
+          return false
+        }
+      }
+    }
+
+    // If tls flag is undefined we suppose that tls is turned on
+    if (flags.tls === false) {
```
tolusha: `return flags.tls !== false` =)

src/tasks/installers/operator.ts
mmorhun: I used info here because it is more like a message. Think of case when real certificate is used. Then we would end up in a situation with unresolvable warning, which is wrong in my opinion.

src/tasks/installers/operator.ts
mmorhun: Yes, possible. But I would like to leave it as is.

PR #606, opened 2020-03-30
fix: Fix e2e tests using self signed certs

flacatus:
### What does this PR do?

This PR enable tls and add custom self signed certificate to e2e tests.

### DIFF
--- a/.ci/self-signed-cert.sh
+++ b/.ci/self-signed-cert.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+#
+# Copyright (c) 2012-2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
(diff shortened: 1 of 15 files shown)


.ci/self-signed-cert.sh
```diff
@@ -0,0 +1,36 @@
...
+openssl req -new -sha256 -key domain.key -subj "/C=US/ST=CK/O=RedHat/CN=${DOMAIN}" -out domain.csr
+
+#Verify Csr
+openssl req -in domain.csr -noout -text
+
+#Generate the certificate using the domain csr and key along with the CA Root key
+openssl x509 -req -in domain.csr -CA rootCA.crt -CAkey rootCA.key -CAcreateserial -out domain.crt -days 500 -sha256
+
+#Verify the certificate's content
+openssl x509 -in domain.crt -text -noout
```
mmorhun: From what I see the certificate misses some constraints/permissions. If for testing purposes it is ok, I do not mind.
I might be wrong, but I am not sure that all tools will accept such certificate without any extension. And if want to use such certificate for, say, selenium tests, we may have problems with it.

flacatus: This tests are not for selenium or any other visual stuff.

flacatus:
[test]

PR #613, opened 2020-04-01
docs: Clarify '--domain' parameter in server:start command

mmorhun:
Signed-off-by: Mykola Morhun <mmorhun@redhat.com>

### What does this PR do?
Adds clarification to `--domain` parameter of `server:start` command.
Adds warning message if `--domain` parameter is used with Openshift family infrastructures.

### What issues does this PR fix or reference?
https://github.com/eclipse/che/issues/14404
https://github.com/eclipse/che/issues/16003

PR #615, opened 2020-04-01
feat: Add domain checks for Kubernetes family infrastructures

mmorhun:

### What does this PR do?
Adds check whether Kubernetes cluster is reachable when installing Che.
The check could be disabled by `--skip-cluster-check` flag.

https://github.com/che-incubator/chectl/pull/613

### DIFF
--- a/src/commands/server/start.ts
+++ b/src/commands/server/start.ts
@@ -140,6 +140,10 @@ export default class Start extends Command {
     'skip-version-check': flags.boolean({
       description: 'Skip minimal versions check.',
       default: false
+    }),
+    'skip-cluster-availability-check': flags.boolean({
+      description: 'Skip cluster availability check. The check is a simple request to ensure the cluster is reachable.',
+      default: false
     })
   }
 
--- a/src/tasks/platforms/common-platform-tasks.ts
+++ b/src/tasks/platforms/common-platform-tasks.ts
@@ -1,0 +1,75 @@
+/*********************************************************************
+ * Copyright (c) 2020 Red Hat, Inc.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ **********************************************************************/
+
+import * as http from 'http'
+import * as https from 'https'
+import * as Listr from 'listr'
+
+export namespace CommonPlatformTasks {
+  /**
+   * Checks whether cluster on which Che is being deployed accessible.
+   * Requires flags.domain to be set.
+   */
+  export function getPingClusterTask(flags: any): Listr.ListrTask {
+    return {
+      title: 'Check if cluster accessible',
+      skip: () => flags['skip-cluster-availability-check'],
+      task: async (_ctx: any, task: any) => {
+        const domain: string = flags.domain
+        if (!domain) {
+          task.title = `${task.title}... "--domain" flag is not set. Cannot check cluster availability.'`
+          return
+        }
+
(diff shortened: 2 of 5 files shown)


src/tasks/platforms/common-platform-tasks.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export namespace CommonPlatformTasks {
+  /**
+   * Checks whether cluster on which Che is being deployed accessible.
+   * Requires flags.domain to be set.
+   */
+  export function getPingClusterTask(flags: any): Listr.ListrTask {
+    return {
+      title: 'Check if cluster accessible',
+      skip: () => flags['skip-cluster-check'],
```
tolusha, line 22: `|| !flags.domain` ?

src/tasks/platforms/common-platform-tasks.ts
```diff
@@ -0,0 +1,56 @@
...
+      task: async (_ctx: any, task: any) => {
+        const domain: string = flags.domain
+        if (!domain) {
+          task.title = `${task.title}... "--domain" flag is not set. Cannot check cluster availability.'`
+          return
+        }
+
+        if (!await checkServer(domain, 80) || !await checkServer(domain, 443)) {
+          throw new Error(`Cannot reach cluster "${domain}". To skip this check add "--skip-cluster-check" flag.`)
```
tolusha, line 31: `neither  ${domain}:80 nor ${domain}:443`

mmorhun: Indeed, thanks

src/tasks/platforms/common-platform-tasks.ts
mmorhun: No, because I want to show warning

src/commands/server/start.ts
```diff
@@ -140,6 +140,10 @@ export default class Start extends Command {
     'skip-version-check': flags.boolean({
       description: 'Skip minimal versions check.',
       default: false
+    }),
+    'skip-cluster-check': flags.boolean({
```
tolusha, line 144: `skip-domain-check`

src/tasks/platforms/common-platform-tasks.ts
tolusha: Instead of
`Cannot reach cluster "${domain}". To skip this check add "--skip-cluster-check" flag.`
it is better to log
`Cannot reach neither '${domain}:80' nor '${domain}:443'. It means the incorrect domain is used. Please, set the correct domain or bypass the check by using '--skip-domain-check' flag.`

mmorhun: I do not think it is a good idea to print domain twice. I would like to avoid it.

src/commands/server/start.ts
mmorhun: `--skip-cluster-availability-check` as we agreed

src/tasks/platforms/common-platform-tasks.ts
```diff
@@ -0,0 +1,75 @@
...
+      skip: () => flags['skip-cluster-availability-check'],
+      task: async (_ctx: any, task: any) => {
+        const domain: string = flags.domain
+        if (!domain) {
+          task.title = `${task.title}... "--domain" flag is not set. Cannot check cluster availability.'`
+          return
+        }
+
+        if (! await checkHttpServer(domain, 80) && ! await checkHttpsServer(domain, 443)) {
+          throw new Error(`Cannot reach cluster at "${domain}". To skip this check add "--skip-cluster-availability-check" flag.`)
```
tolusha, line 32: To skip -> To bypass

PR #628, opened 2020-04-08
fix: Fix e2e tests by adding certificates

flacatus:
### What does this PR do?
This PR add certificates to e2e tests to tests chectl startup on minikube and minishift.


flacatus:
[test]

PR #632, opened 2020-04-12
fix: Reuse domain for server:update command

tolusha:

### What does this PR do?
Allow to use domain flag for server:update command

### DIFF
--- a/src/commands/server/update.ts
+++ b/src/commands/server/update.ts
@@ -16,12 +16,14 @@ import * as Listr from 'listr'
 import * as notifier from 'node-notifier'
 import * as path from 'path'
 
+import { KubeHelper } from '../../api/kube'
 import { cheDeployment, cheNamespace, listrRenderer } from '../../common-flags'
-import { DEFAULT_CHE_OPERATOR_IMAGE } from '../../constants'
+import { CHE_CLUSTER_CR_NAME, DEFAULT_CHE_OPERATOR_IMAGE } from '../../constants'
 import { CheTasks } from '../../tasks/che'
 import { InstallerTasks } from '../../tasks/installers/installer'
 import { ApiTasks } from '../../tasks/platforms/api'
 import { PlatformTasks } from '../../tasks/platforms/platform'
+import { isKubernetesPlatformFamily } from '../../util'
 
 export default class Update extends Command {
   static description = 'update Eclipse Che server'
@@ -125,6 +127,9 @@ export default class Update extends Command {
       if (!ctx.isCheDeployed) {
         this.error('Eclipse Che deployment is not found. Use `chectl server:start` to initiate new deployment.')
       } else {
+        if (isKubernetesPlatformFamily(flags.platform!)) {
+          await this.setDomainFlag(flags)
+        }
         await platformCheckTasks.run(ctx)
 
         await preUpdateTasks.run(ctx)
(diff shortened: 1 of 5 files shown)


tolusha:
[test]

README.md
```diff
@@ -385,33 +385,43 @@ USAGE
...
 OPTIONS
-  -a, --installer=helm|operator|minishift-addon                                Installer type
-  -h, --help                                                                   show CLI help
+  -a, --installer=helm|operator|minishift-addon
+      Installer type
 
-  -n, --chenamespace=chenamespace                                              [default: che] Kubernetes namespace where
-                                                                               Eclipse Che server is supposed to be
-                                                                               deployed
+  -b, --domain=domain
```
mmorhun, line 391: To me, using of `-b` short key is weird. Could you please explain logic behind it? Why not `-d` for **d**omain?

tolusha: I have no idea.

sleshchenko: I suppose from the very beginning short aliases where generated in alphabetic order.
+1 to start cleaning them up, start deprecation of weird ones and introduce new more logical ones.

src/commands/server/update.ts
```diff
@@ -38,6 +38,16 @@ export default class Update extends Command {
       description: 'Type of Kubernetes platform. Valid values are \"minikube\", \"minishift\", \"k8s (for kubernetes)\", \"openshift\", \"crc (for CodeReady Containers)\", \"microk8s\".',
       options: ['minikube', 'minishift', 'k8s', 'openshift', 'microk8s', 'docker-desktop', 'crc'],
     }),
+    domain: string({
```
sleshchenko, line 41: Can we make UX even smoother? 
I suppose now doman parameter is required and installation will be broker if an admin miss it during server:update for K8s, right?
If so - could we supply this option only when user needs to change it, otherwise - the initially configure value could be untouched?

tolusha: > If so - could we supply this option only when user needs to change it

I don't understand

src/commands/server/update.ts
```diff
@@ -1,4 +1,4 @@
-/*********************************************************************
+1/*********************************************************************
```
sleshchenko, line 1:
```suggestion
/*********************************************************************
```

src/commands/server/update.ts
```diff
@@ -151,4 +156,12 @@ export default class Update extends Command {
 
     this.exit(0)
   }
+
+  async setDomainFlag(flags: any): Promise<void> {
+    const kubeHelper = new KubeHelper()
+    const cheCluster = await kubeHelper.getCheCluster('eclipse-che', flags.chenamespace)
```
sleshchenko, line 162: `eclipse-che` must be extracted into constant to make it possible easier customize in crwctl

sleshchenko: It does not seem to be a universal way to extract domain for all installer, it would be better to make it a part of `installers/operator.ts`

tolusha: It is related to platform, not to installer type

sleshchenko: Really?
I believe it's related to Operator installer
Helm chart does not have such Custom Resource
But I may be wrong

tolusha: It is used to deploy Eclipse Che
https://github.com/che-incubator/chectl/blob/ab/domainupdate/src/tasks/installers/helm.ts#L331

sleshchenko: Ok, you mean domain flag - and I agree with that, but actually I mean using CustomResource Che Cluster...
The way how the domain is evaluated on already deployed Che - depends on the installer.
For Operator - it's CustomResource CheCluster
For Helm Chart - it's directly Che Config Map, right?

tolusha: Right, but we don't support update in case of helm installer.

PR #639, opened 2020-04-14
feat: Improve workspace:inject command

tolusha:

### What does this PR do?
Allows to inject configuration into a workspace that is running in non che namespace.
The workspace namespace is retrieved from the workspace configuration. 

https://github.com/eclipse/che/issues/16019


src/commands/workspace/inject.ts
```diff
@@ -31,59 +33,78 @@ export default class Inject extends Command {
...
+      if (!workspaceId) {
+        const workspaces = await cheHelper.getWorkspaces(cheURL, flags['access-token'])
+        const runningWorkspaces = workspaces.filter(w => w.status === 'RUNNING')
+        if (runningWorkspaces.length === 1) {
+          workspaceId = runningWorkspaces[0].id
+          workspaceNamespace = runningWorkspaces[0].attributes.infrastructureNamespace
+        } else if (runningWorkspaces.length === 0) {
+          cli.error('There are no running workspaces. Please start workspace first.')
+        } else {
+          cli.error('There are more than 1 running workspaces. Please, specify the workspace id by providing \'--workspace\' flag.\nSee more help with --help.')
```
mmorhun, line 92: What about `See more information by using --help flag`?

src/api/che.ts
```diff
@@ -437,10 +437,24 @@ export class CheHelper {
       () => { })
   }
 
+  async getWorkspaces(cheURL: string, accessToken?: string): Promise<any[]> {
+    const all: any[] = []
+    const maxItems = 30
+    let skipCount = 0
+
+    do {
```
AndrienkoAleksandr, line 445: Is it make sense do cycle if default maxItems is 30? https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java#L344

src/api/che.ts
```diff
@@ -437,10 +437,24 @@ export class CheHelper {
       () => { })
   }
 
+  async getWorkspaces(cheURL: string, accessToken?: string): Promise<any[]> {
```
AndrienkoAleksandr, line 440: Maybe getAllWorkspaces?

PR #647, opened 2020-04-16
feat: Print the current context when chectl is run

tolusha:

### What does this PR do?
Prints the current context when chectl is run

https://github.com/eclipse/che/issues/15741

### DIFF
--- a/src/api/kube.ts
+++ b/src/api/kube.ts
@@ -28,19 +28,26 @@ import { V1alpha2Certificate } from './typings/cert-manager'
 const AWAIT_TIMEOUT_S = 30
 
 export class KubeHelper {
-  kc = new KubeConfig()
-  portForwardHelper = new PortForward(this.kc, true)
-  logHelper = new Log(this.kc)
+  static readonly kc = KubeHelper.initializeKubeConfig()
+  static initializeKubeConfig(): KubeConfig {
+    const kc = new KubeConfig()
+    if (process.env.KUBECONFIG) {
+      kc.loadFromFile(process.env.KUBECONFIG)
+      cli.info(`kubeconfig => loaded from the file ${process.env.KUBECONFIG}`)
+    } else {
+      kc.loadFromDefault()
+      cli.info('kubeconfig => loaded default configuration.')
+    }
+    return kc
+  }
+
+  portForwardHelper = new PortForward(KubeHelper.kc, true)
+  logHelper = new Log(KubeHelper.kc)
 
   podWaitTimeout: number
   podReadyTimeout: number
 
-  constructor(flags?: any, context?: string) {
-    if (!context) {
-      this.kc.loadFromDefault()
-    } else {
-      this.kc.loadFromString(context)
-    }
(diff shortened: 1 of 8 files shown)


tolusha:
[test]

src/api/kube.ts
```diff
@@ -54,7 +61,7 @@ export class KubeHelper {
   }
 
   async deleteAllServices(namespace = '') {
-    const k8sApi = this.kc.makeApiClient(CoreV1Api)
+    const k8sApi = KubeHelper.KUBE_CONFIG.makeApiClient(CoreV1Api)
```
mmorhun, line 64: What if we need to have different configs? For example make queries from different accounts?

tolusha: I have not faced such a case

tolusha:
After digging into the code I figured out that `KubeConfig` already takes into account `KUBECONFIG` env var

```javascript
        if (process.env.KUBECONFIG && process.env.KUBECONFIG.length > 0) {
            const files = process.env.KUBECONFIG.split(path.delimiter);
            this.loadFromFile(files[0]);
            for (let i = 1; i < files.length; i++) {
                const kc = new KubeConfig();
                kc.loadFromFile(files[i]);
                this.mergeConfig(kc);
            }
            return;
        }
```

src/api/kube.ts
mmorhun: But we may face it one day...

tolusha: Could you elaborate more?

tolusha:
[test]

src/api/kube.ts
```diff
@@ -28,19 +28,21 @@ import { V1alpha2Certificate } from './typings/cert-manager'
 const AWAIT_TIMEOUT_S = 30
 
 export class KubeHelper {
-  kc = new KubeConfig()
-  portForwardHelper = new PortForward(this.kc, true)
-  logHelper = new Log(this.kc)
+  public static readonly KUBE_CONFIG = KubeHelper.initializeKubeConfig()
```
mmorhun, line 31: I am not keen on the idea to make kubeconfig static.

tolusha: Why?

mmorhun: `KubeHelper` is helper class which is used to communicate with Kubernetes API. By making kubeconfig static you basically make whole class static, so it becomes more a namespace than a class... It is my vision...

src/api/kube.ts
mmorhun: I mean not this specific line of code, I am talking about more general stuff. Maybe we do not need it at all (as we don't right now), but static kubeconfig makes chectl somewhat limited.

tolusha: It is absolutely not true, it brings consistency. Image the case when `chectl` is working and meanwhile user changed the context? Then a new instance of KubeHelper will use a kubeconfig with a different context that might lead to errors.

tolusha:
[test]

PR #656, opened 2020-04-20
feat: Add dashboard:open command

tolusha:

### What does this PR do?
Add `dashboard:open`  command to open Che dashboard

https://github.com/eclipse/che/issues/15003

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ USAGE
 # Commands
 <!-- commands -->
 * [`chectl autocomplete [SHELL]`](#chectl-autocomplete-shell)
+* [`chectl dashboard:open`](#chectl-dashboardopen)
 * [`chectl devfile:generate`](#chectl-devfilegenerate)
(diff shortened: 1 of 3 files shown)


l0rd:
It would be better to call the command `dashboard:open` or `dashboard:info`.

src/commands/server/dashboard.ts
```diff
@@ -0,0 +1,52 @@
...
+        cli.url(dashboardUrl, dashboardUrl)
+      } else {
+        cli.info(`Opening ... ${dashboardUrl}`)
+        await cli.open(dashboardUrl)
+      }
+    } catch (error) {
+      this.error(error)
+    }
+
+    notifier.notify({
```
mmorhun, line 45: won't it be executed even in case of error form catch block above?

tolusha: No, `this.error` prints the error and terminates the program.

PR #669, opened 2020-04-24
feat: Add cacert:export command to get self-signed Che CA certificate

mmorhun:

### What does this PR do?
Adds a new command `cacert:export` which retrieves self-signed Che CA certificate and saves it to specified file.
Note, it works only for self-signed certificates.

Tested on `minikube`, `crc` and `openshift`.

https://github.com/eclipse/che/issues/16278

### DIFF
--- a/src/api/che.ts
+++ b/src/api/che.ts
@@ -114,14 +114,14 @@ export class CheHelper {
   async retrieveEclipseCheCaCert(cheNamespace: string): Promise<string> {
     const cheCaSecret = await this.kube.getSecret(CHE_ROOT_CA_SECRET_NAME, cheNamespace)
     if (!cheCaSecret) {
-      throw new Error('Local Che CA self-signed certificate not found. Are you using self-signed certificate?')
+      throw new Error('Che CA self-signed certificate not found. Are you using self-signed certificate?')
     }
 
     if (cheCaSecret.data && cheCaSecret.data['ca.crt']) {
       return Buffer.from(cheCaSecret.data['ca.crt'], 'base64').toString('ascii')
     }
 
-    throw new Error(`Secret "${CHE_ROOT_CA_SECRET_NAME}" has invalid format.`)
+    throw new Error(`Secret "${CHE_ROOT_CA_SECRET_NAME}" has invalid format: "ca.crt" key not found in data.`)
   }
 
   async cheK8sURL(namespace = ''): Promise<string> {
--- a/src/api/che.ts
+++ b/src/api/che.ts
@@ -20,6 +20,7 @@ import * as yaml from 'js-yaml'
 import * as path from 'path'
 
 import { OpenShiftHelper } from '../api/openshift'
+import { CHE_ROOT_CA_SECRET_NAME } from '../constants'
 
 import { Devfile } from './devfile'
 import { KubeHelper } from './kube'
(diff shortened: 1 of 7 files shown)


src/commands/server/certificate.ts
```diff
@@ -0,0 +1,105 @@
...
+    destination: string({
+      char: 'd',
+      description: `Destination where to store Che certificate.
+                    If the destination is a file (might not exist), then the certificate will be saved there in PEM format.
+                    If the destination is a directory, then ${DEFAULT_CA_CERT_FILE_NAME} file will be created there with Che certificate in PEM format.
+                    If this option is ommited, then Che certificate will be stored in user's home directory as ${DEFAULT_CA_CERT_FILE_NAME}`,
+      env: 'CHE_CA_CERT_LOCATION',
+      default: '~'
+    }),
+    'make-path': boolean({
```
tolusha, line 45: Not sure it is necessary at all

mmorhun: It might be useful in some cases. Anyway, it is implemented and works fine, see no point in removing...

AndrienkoAleksandr: Maybe we should make path by default without extra flags?

src/api/che.ts
```diff
@@ -107,6 +108,22 @@ export class CheHelper {
...
+    const cheCaSecret = await this.kube.getSecret(CHE_ROOT_CA_SECRET_NAME, cheNamespace)
+    if (!cheCaSecret) {
+      throw new Error('Local Che CA self-signed certificate not found. Are you using self-signed certificate?')
+    }
+
+    if (cheCaSecret.data && cheCaSecret.data['ca.crt']) {
+      return Buffer.from(cheCaSecret.data['ca.crt'], 'base64').toString('ascii')
+    }
+
+    throw new Error(`Secret "${CHE_ROOT_CA_SECRET_NAME}" has invalid format.`)
```
tolusha, line 124: Add pls.
`ca.crt key not found.`

src/commands/cacert/export.ts
```diff
@@ -0,0 +1,88 @@
...
+      options: ['minikube', 'minishift', 'k8s', 'openshift', 'microk8s', 'docker-desktop', 'crc'],
+    }),
+    destination: string({
+      char: 'd',
+      description: `Destination where to store Che certificate.
+                    If the destination is a file (might not exist), then the certificate will be saved there in PEM format.
+                    If the destination is a directory, then ${DEFAULT_CA_CERT_FILE_NAME} file will be created there with Che certificate in PEM format.
+                    If this option is ommited, then Che certificate will be stored in user's home directory as ${DEFAULT_CA_CERT_FILE_NAME}`,
+      env: 'CHE_CA_CERT_LOCATION',
+      default: '~'
```
tolusha, line 44: That is unix specific.
Let's have empty default.

src/commands/cacert/export.ts
```diff
@@ -0,0 +1,88 @@
...
+    } catch (error) {
+      this.error(error)
+    }
+  }
+
+  /**
+   * Handles certificate target location and returns string which points to the target file.
+   */
+  private prepareTarget(destinaton: string): string {
+    if (destinaton === '~') {
```
tolusha, line 77: if (!destination)

src/commands/cacert/export.ts
```diff
@@ -0,0 +1,88 @@
...
+      return path.join(os.homedir(), DEFAULT_CA_CERT_FILE_NAME)
+    }
+
+    if (fs.existsSync(destinaton)) {
+      return fs.lstatSync(destinaton).isDirectory() ? path.join(destinaton, DEFAULT_CA_CERT_FILE_NAME) : destinaton
+    }
+
+    throw new Error('Given certificate path doesn\'t exist.')
```
tolusha, line 85: Which path? Add path into the error

src/commands/cacert/export.ts
mmorhun: At the end it isn't. But agree, to avoid confusion.

src/commands/cacert/export.ts
mmorhun: Doesn't work with `~` =)

src/commands/cacert/export.ts
mmorhun: Make sense

src/commands/cacert/export.ts
tolusha: if (!destination) => home dir by default
else path.resolve()

PR #683, opened 2020-05-03
fix: Check amount users when OAuth is enabled.

AndrienkoAleksandr:
### What does this PR do?
Check amount users when OAuth is enabled.

https://github.com/eclipse/che/issues/16629

Signed-off-by: Oleksandr Andriienko <oandriie@redhat.com>


tolusha:
@AndrienkoAleksandr 
Please provide steps to test the PR.

AndrienkoAleksandr:
> Please provide CR to test the PR.

Hello, we can use flag  --os-oauth:

```
./run server:start -m -p crc -n moon50  --os-oauth
```

src/constants.ts
```diff
@@ -32,3 +32,5 @@ export const defaultOLMKubernetesNamespace = 'olm'
 export const DOCS_LINK_INSTALL_TLS_WITH_SELF_SIGNED_CERT = 'https://www.eclipse.org/che/docs/che-7/installing-che-in-tls-mode-with-self-signed-certificates/'
 export const DOCS_LINK_IMPORT_CA_CERT_INTO_BROWSER = 'https://www.eclipse.org/che/docs/che-7/installing-che-in-tls-mode-with-self-signed-certificates/#using-che-with-tls_installing-che-in-tls-mode-with-self-signed-certificates'
 export const DOCS_LINK_AUTH_TO_CHE_SERVER_VIA_OPENID = ' https://www.eclipse.org/che/docs/che-7/authenticating-users/#authenticating-to-the-che-server-using-openid_authenticating-to-the-che-server'
+export const HOW_TO_CREATE_USER_OS4 = 'https://docs.openshift.com/container-platform/4.1/authentication/understanding-identity-provider.html#identity-provider-overview_understanding-identity-provider'
```
mmorhun, line 35: Please add `DOCS_LINK_` prefix.

src/tasks/platforms/crc.ts
```diff
@@ -54,6 +57,16 @@ export class CRCHelper {
       },
       VersionHelper.getOpenShiftCheckVersionTask(flags),
       VersionHelper.getK8sCheckVersionTask(flags),
+      {
+        title: 'Verify amount "crc" users',
```
sleshchenko, line 61: Do we really care about users number?
Would it better to just `Verify the existence of users?`

src/tasks/platforms/crc.ts
```diff
@@ -54,6 +57,16 @@ export class CRCHelper {
...
+        enabled: () => flags['os-oauth'],
+        task: async (_ctx: any, task: any) => {
+          if (await kube.getAmoutUsers() === 0) {
+            command.error(`No real user exists in the "crc" cluster. Either disable OpenShift OAuth integration("os-oauth" flag) or add at least one user (details in the Help link): "${HOW_TO_CREATE_USER_OS4}"`)
```
sleshchenko, line 65: I don't really understand why we have such a requirement for OpenShift Cluster to have a real user...
I could imagine a new OS cluster where admin set up everything before anybody use it:
- add OAuth, like htpasswd;
- install Che (and now it's not possible, because admin except setting up OAuth on cluster must ask anybody or register with set up OAuth on the cluster)...

Why we can't just tell a warning after Che is installed - `Hey, you installed Che but it's not possible to use kube:admin on the Che. Please use some other logging methods`?

l0rd: @AndrienkoAleksandr @sleshchenko we [recently had some feedback](https://github.com/eclipse/che/issues/14013#issuecomment-622386571) about that. The use case for a cluster with only kubeadmin is pretty common. That is something we need to support better.

AndrienkoAleksandr: > Why we can't just tell a warning after Che is installed

Sorry, but Che won't be installed. Installation will fail.  We should notify user as fast, as we can about an issues, that's why we are using pre-flight checks. Otherwise user will wait 5-10 minutes to get broken installation.

sleshchenko: I know about that check on the Operator side but could you elaborate more why Che can't be installed without a real user?

AndrienkoAleksandr:
Closed on favour https://github.com/che-incubator/chectl/pull/692

PR #689, opened 2020-05-06
fix: Apply CR patch after all manipulations

mmorhun:

### What does this PR do?
Move applying CR patch for operator installer after predefined CR corrections on chectl side.


flacatus:
checking

PR #692, opened 2020-05-07
fix: Disable oauth if there are no identity providers.

AndrienkoAleksandr:
### What does this PR do?
Disable oauth if there are no identity providers.

https://github.com/eclipse/che/issues/16836

Signed-off-by: Oleksandr Andriienko <oandriie@redhat.com>


tolusha:
Have a look at errors:
`ERROR: 21:1  ordered-imports  Import sources within a group must be alphabetized.`
run `yarn test` to figure out all errors.

src/tasks/platforms/common-platform-tasks.ts
```diff
@@ -72,4 +75,19 @@ export namespace CommonPlatformTasks {
     })
   }
 
+  export function oAuthProvidersExists(flags: any): Listr.ListrTask {
+    let kube = new KubeHelper(flags)
+    return {
+      title: 'Verify the existence of identity providers.',
+      enabled: () => isOpenshiftPlatformFamily(flags.platform) && flags['os-oauth'],
```
tolusha, line 82: I think checking `flags['os-oauth']` is not sufficient. 
Use can set this flag in CR also.  We need something similar to
https://github.com/che-incubator/chectl/blob/master/src/commands/server/start.ts#L219

tolusha:
[test]

src/tasks/platforms/common-platform-tasks.ts
```diff
@@ -72,4 +78,47 @@ export namespace CommonPlatformTasks {
...
+  /**
+   * Checks if Openshift oAuth is disabled via operator custom resource.
+   * Returns true if Openshift oAuth is enabled (or omitted) and false if it is explicitly disabled.
+   */
+  async function isOAuthEnabled(flags: any): Promise<boolean> {
+    if (flags['che-operator-cr-yaml']) {
+      const cheOperatorCrYamlPath = flags['che-operator-cr-yaml']
+      if (fs.existsSync(cheOperatorCrYamlPath)) {
+        const cr = yaml.safeLoad(fs.readFileSync(cheOperatorCrYamlPath).toString())
+        if (cr && cr.spec && cr.spec.auth && cr.spec.auth.openShiftoAuth) {
```
tolusha, line 106: I am doubt if `&& cr.spec.auth.openShiftoAuth` is correct check.
What if user defined `cr.spec.auth.openShiftoAuth: false` ?

src/tasks/installers/common-tasks.ts
```diff
@@ -8,6 +8,7 @@
  * SPDX-License-Identifier: EPL-2.0
  **********************************************************************/
 
+import ansi = require('ansi-colors')
```
mmorhun, line 11: Can we use `import * as ansi from 'ansi-colors'` ?

AndrienkoAleksandr: 'ansi-colors' lib hasn't got d.ts file.

src/tasks/platforms/common-platform-tasks.ts
```diff
@@ -72,4 +79,54 @@ export namespace CommonPlatformTasks {
...
+    let kube = new KubeHelper(flags)
+    return {
+      title: 'Verify Openshift oauth.',
+      enabled: () => isOpenshiftPlatformFamily(flags.platform) && isOAuthEnabled(flags),
+      task: async (ctx: any, task: any) => {
+        if (await kube.isOpenShift4()) {
+          const providers = await kube.getOpenshiftAuthProviders()
+          if (!providers || providers.length === 0) {
+            ctx.highlightedMessages.push(`❗ ${ansi.yellow('[WARNING]')} 'os-oauth' flag was disabled, because Openshift oauth hasn't got any identity providers. ${DOCS_LINK_HOW_TO_ADD_IDENTITY_PROVIDER_OS4}`)
+            flags['os-oauth'] = false
```
tolusha, line 92: Basically here
https://github.com/che-incubator/chectl/blob/master/src/api/kube.ts#L1133
we ignore flags if provided cr is used ^(

AndrienkoAleksandr: Good catch.

PR #694, opened 2020-05-12
feat: Use `rejectUnauthorized: false`

tolusha:

### What does this PR do?
Don't check che-server identity.

https://issues.redhat.com/browse/CRW-864

### DIFF
--- a/src/api/che.ts
+++ b/src/api/che.ts
@@ -8,11 +8,12 @@
  * SPDX-License-Identifier: EPL-2.0
  **********************************************************************/
 import { CoreV1Api, KubeConfig, V1Pod, Watch } from '@kubernetes/client-node'
-import axios from 'axios'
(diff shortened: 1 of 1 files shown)


tolusha:
@mmorhun 
I am wondering if it is better to introduce `--cacert` instead.

mmorhun:
@tolusha the right way is to pass CA certificate for Axios instance. However, requiring from user to manually specify the CA certificate is not user friendly way (we are working on TLS automation right now). So I may suggest to leave current approach with `rejectUnauthorized: false` as this is a backport and in future rethink our approach. Created a separate [issue](https://github.com/eclipse/che/issues/16906) as the task is out of scope of this PR.

nickboldt:
3 tests have failed. Do we care? If not, please merge. If yes, please fix / retrigger.

PR #698, opened 2020-05-13
fix: Update che-operator and che versions

tolusha:

### What does this PR do?
Update che-operator and che versions


tolusha:
[test]

PR #700, opened 2020-05-14
fix: Use 'openshift-marketplace' for OpenShift platform familiy

tolusha:

### What does this PR do?
Using 'openshift-marketplace' for `catalogSourceNamespace` for OpenShift platform familiy

https://github.com/eclipse/che/issues/16913

### DIFF
--- a/src/tasks/installers/installer.ts
+++ b/src/tasks/installers/installer.ts
@@ -89,7 +89,7 @@ export class InstallerTasks {
 
     // let task: Listr.ListrTask
 
-    if (flags.installer === 'operator' || flags.installer === 'olm') {
+    if (flags.installer === 'operator') {
       title = '🏃‍  Running the Eclipse Che operator'
       task = () => {
         // The operator installs Eclipse Che in multiuser mode by default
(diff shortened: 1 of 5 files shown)


src/tasks/installers/installer.ts
```diff
@@ -89,7 +89,7 @@ export class InstallerTasks {
 
     // let task: Listr.ListrTask
 
-    if (flags.installer === 'operator') {
+    if (flags.installer === 'operator' || flags.installer === 'olm') {
```
AndrienkoAleksandr, line 92: Sorry, but we have flag 'flags.installer === 'olm'' in the line 103. And with current change will be printed `'🏃‍  Running the Eclipse Che operator'`

tolusha: Right. Thank you

src/tasks/installers/olm.ts
```diff
@@ -52,7 +52,7 @@ export class OLMTasks {
         title: 'Configure context information',
         task: async (ctx: any, task: any) => {
           // Todo: should we do check for installer openshift? flags.platform === 'crc' || flags.platform === 'openshift'
```
AndrienkoAleksandr, line 54: Remove todo. You fixed it :)

PR #701, opened 2020-05-14
fix: Detect olm installer type for server:update command

tolusha:

### What does this PR do?
Check if subscription exists and automatically set installer type to `olm` if it wasn't specified.

https://github.com/eclipse/che/issues/16861

### DIFF
--- a/src/commands/server/start.ts
+++ b/src/commands/server/start.ts
@@ -10,6 +10,7 @@
 
 import { Command, flags } from '@oclif/command'
 import { boolean, string } from '@oclif/parser/lib/flags'
+import { cli } from 'cli-ux'
 import * as fs from 'fs-extra'
 import * as yaml from 'js-yaml'
 import * as Listr from 'listr'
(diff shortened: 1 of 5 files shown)


src/commands/server/start.ts
```diff
@@ -417,4 +419,19 @@ export default class Start extends Command {
...
   }
+
+  /**
+   * Sets default installer which is `olm` for OpenShift 4 with stable version of chectl
+   * and `operator` for other cases.
+   */
+  async setDefaultInstaller(flags: any): Promise<void> {
+    const cheVersion = DEFAULT_CHE_OPERATOR_IMAGE.split(':')[1]
+    const kubeHelper = new KubeHelper(flags)
+    if (flags.platform === 'openshift' && await kubeHelper.isOpenShift4() && cheVersion !== 'nightly' && cheVersion !== 'latest') {
```
AndrienkoAleksandr, line 430: what  about `crc` ?

tolusha: It is just for openshift

PR #714, opened 2020-05-19
fix(messages): Improve warn messages

tolusha:

### What does this PR do?
Improves some warn messages

https://github.com/eclipse/che/issues/16968

### DIFF
--- a/src/tasks/installers/helm.ts
+++ b/src/tasks/installers/helm.ts
@@ -34,7 +34,7 @@ export class HelmTasks {
    * Returns list of tasks which perform preflight platform checks.
    */
   startTasks(flags: any, command: Command): Listr {
-    if (isStableVersion()) {
+    if (isStableVersion(flags)) {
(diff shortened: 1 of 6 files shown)


src/tasks/installers/helm.ts
```diff
@@ -34,7 +34,9 @@ export class HelmTasks {
    * Returns list of tasks which perform preflight platform checks.
    */
   startTasks(flags: any, command: Command): Listr {
-    command.warn('You can also use features rich \'OLM\' installer to deploy Eclipse Che.')
+    if (isStableVersion(flags)) {
+      command.warn('Consider using the more reliable \'OLM\' installer when deploying a stable release of Eclipse Che (--installer=olm).')
```
mmorhun, line 38: I don't see why OLM is
 > more reliable

tolusha: - no cluster admin rights
- OpenShift manages installation

mmorhun: Maybe then `easier to use`?

tolusha: Not sure, see the related issue.

PR #715, opened 2020-05-19
fix: Change OS and nodejs version

flacatus:
Signed-off-by: flacatus <flacatus@redhat.com>

### What does this PR do?
This PR change os jobs from OSX to linux and add the latest stable version of nodejs.

https://github.com/eclipse/che/issues/16801

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 jobs:
   include:
-   - os: osx
+   - os: linux
 language: node_js
-node_js: '10'
+node_js: '12'
 before_deploy:
 - |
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,8 +1,8 @@
 jobs:
   include:
-   - os: osx
+   - os: linux
 language: node_js
-node_js: '10'
+node_js: '12'
```
tolusha, line 5: Not sure. Why do we need this?

flacatus: I just change to LTS recommended here https://nodejs.org/en/ . But we can still use 10.X version

PR #718, opened 2020-05-19
fix: Revert travis ci changes

flacatus:
Signed-off-by: flacatus <flacatus@redhat.com>

### What does this PR do?
Revert latest changes of travis CI.


benoitf:
why there is only s390x arch ?

PR #729, opened 2020-05-25
fix: Improve getPodReadyConditionStatus

tolusha:

### What does this PR do?
Sometimes pod `status` might not contain `condition` of type `Ready`.
It is better to return `underfine` for getPodReadyConditionStatus and allows other function to wait and get `condition` in the next request.

https://github.com/eclipse/che/issues/17001

### DIFF
--- a/src/api/kube.ts
+++ b/src/api/kube.ts
@@ -558,7 +558,7 @@ export class KubeHelper {
     return res.body.items[0].status.phase
   }
 
-  async getPodReadyConditionStatus(selector: string, namespace = ''): Promise<string> {
+  async getPodReadyConditionStatus(selector: string, namespace = ''): Promise<string | undefined> {
     const k8sCoreApi = KubeHelper.KUBE_CONFIG.makeApiClient(CoreV1Api)
     let res
     try {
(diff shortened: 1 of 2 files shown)


src/api/kube.ts
```diff
@@ -568,24 +568,20 @@ export class KubeHelper {
...
-
-    if (!res.body.items[0].status) {
-      throw new Error(`Get pods by selector "${selector}" returned a pod with an invalid state`)
+      throw new Error(`Get pods by selector "${selector}" returned ${res.body.items.length} pods (1 was expected).`)
     }
 
-    if (!res.body.items[0].status.conditions || !(res.body.items[0].status.conditions.length > 0)) {
-      throw new Error(`Get pods by selector "${selector}" returned a pod with an invalid status.conditions`)
+    if (!res.body.items[0].status || !res.body.items[0].status.conditions || !(res.body.items[0].status.conditions.length > 0)) {
+      return
```
mmorhun, line 584: please return `undefined` explicitly

src/api/kube.ts
```diff
@@ -568,24 +568,20 @@ export class KubeHelper {
     }
 
     if (!res || !res.body || !res.body.items) {
-      throw new Error(`Get pods by selector "${selector}" returned an invalid response`)
+      throw new Error(`Get pods by selector "${selector}" returned an invalid response.`)
     }
 
+    // No pods found by the specified selector. So, it's not ready.
```
mmorhun, line 574: Are you sure that this is the right place for this comment?

tolusha: What do you mean by that?

src/api/kube.ts
tolusha: I can't otherwise we will have lint errors

src/api/kube.ts
mmorhun: To my mind this comment should be inside `if` block. Because in the place where this comment is now it is not true that "No pods found by the specified selector".

src/api/kube.ts
mmorhun: :disappointed:

src/api/kube.ts
tolusha: Fixed

PR #752, opened 2020-06-22
fix: Add workaround for Minikube storage provisioning bug

mmorhun:

### What does this PR do?
Minikube starting from version `1.9` has [a bug](https://github.com/kubernetes/minikube/issues/7218) which prevents Che and some other products from successful deployment. The root cause is the outdated storage-provisioner image and updating it to the latest version fixes the problem, but it also requires to add some permissions to the storage-provisioner cluster role. That is done in this PR on platform preparation phase.

However, we should update range of Minikube versions for which this workaround is applied when Minikube with the fix released.

Tested with Minikube 1.11.0

https://github.com/eclipse/che/issues/17142

### DIFF
--- a/src/tasks/platforms/minikube.ts
+++ b/src/tasks/platforms/minikube.ts
@@ -13,6 +13,7 @@ import * as commandExists from 'command-exists'
 import * as execa from 'execa'
 import * as Listr from 'listr'
 
+import { KubeHelper } from '../../api/kube'
 import { VersionHelper } from '../../api/version'
 
 import { CommonPlatformTasks } from './common-platform-tasks'
@@ -22,6 +23,7 @@ export class MinikubeTasks {
    * Returns tasks list which perform preflight platform checks.
    */
   preflightCheckTasks(flags: any, command: Command): Listr {
+    const kube = new KubeHelper(flags)
     return new Listr([
       {
         title: 'Verify if kubectl is installed',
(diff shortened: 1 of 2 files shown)


flacatus:
[test]

src/tasks/platforms/minikube.ts
```diff
@@ -79,6 +81,49 @@ export class MinikubeTasks {
...
+          task.title = `${task.title}... ${version}`
+        }
+      },
+      {
+        // Starting from Minikube 1.9 there is a bug with storage provisioner which prevents Che from successful deployment.
+        // For more details see https://github.com/kubernetes/minikube/issues/7218
+        // To workaround the bug, it is required to patch storage provisioner as well as its permissions.
+        title: 'Patch minikube storage',
+        enabled: ctx => ctx.minikubeVersionMajor && ctx.minikubeVersionMinor &&
+          ctx.minikubeVersionMajor === 1 && ctx.minikubeVersionMinor >= 9 && ctx.minikubeVersionMinor <= 11,
```
l0rd, line 102: Why do we need to check that the version is <= 11? I mean I would set the upper bound only when we know wich version of minikube will contain the fix. Otherwise we will get a blocker as soon as 1.12 is out.

src/tasks/platforms/minikube.ts
```diff
@@ -79,6 +81,49 @@ export class MinikubeTasks {
...
+        task: async (_ctx: any, task: any) => {
+          // Patch storage provisioner pod to the latest version
+          const storageProvisionerImagePatch = {
+            apiVersion: 'v1',
+            kind: 'Pod',
+            spec: {
+              containers: [
+                { name: 'storage-provisioner', image: 'gcr.io/k8s-minikube/storage-provisioner:latest' }
```
l0rd, line 110: Instead of `latest` here we should use use the digest (or a version) to avoid that an update to `latest` breaks things.

mmorhun: ok, I will use digest as there is no tag (if don't count `latest`) for the images above `v1.8.1` which is 2 years old.

src/tasks/platforms/minikube.ts
mmorhun: It is sort of a template when the bug is fixed. Seems there are some activity in the Minikube issue, so I hoped that it will be fixed soon. But agree, we'll add stop condition when Minikube fix arrives.

AndrienkoAleksandr:
Tested with minikube 1.11 installation works fine. Workspaces starts too.

flacatus:
Tested as well in my local machine with latest minikube version.
I think we can use latest minikube version here:
https://github.com/che-incubator/chectl/blob/master/.ci/start-minikube.sh#L14-L15

AndrienkoAleksandr:
> I think we can use latest minikube version here:

Then we need set up argument `--driver=virtualbox` for `minikube start` command, otherwise minikube will use docker...

flacatus:
> > I think we can use latest minikube version here:
> 
> Then we need set up argument `--driver=virtualbox` for `minikube start` command, otherwise minikube will use docker...

In ci.centos unfortunetly we cannot use virtualbox or kvm2 drivers only --driver=none :(

PR #761, opened 2020-06-26
fix: explain that delete only works on stopped...

nickboldt:
explain that delete only works on stopped workspaces - #17256 CRW-1027

Change-Id: I38666faab7593c6a79cb602b258e97be36f6ab28


AndrienkoAleksandr:
[test]

PR #789, opened 2020-07-13
fix: Don't throw an error when there is no checlusters crd

tolusha:

### What does this PR do?
`server:delete` command should not throw an error trying to remove `checlusters` CR when Eclipse Che has been deploying using helm.

https://github.com/eclipse/che/issues/17375


flacatus:
[test]

PR #812, opened 2020-08-12
fix(minikube): do not enable ingress addon if already enabled

benoitf:
### What does this PR do?
new output of addons list command is returning pipe characters instead of : as separator thus the check is failing
use new json output to make it more reliable (if json output is available)

Include tests as well

new output on recent minikube versions:
image

old output:
image

https://github.com/eclipse/che/issues/17624

Change-Id: Ifd0c8f716c42ecc2c412ba8d6964e3e22b8d93f5


benoitf:
I've fixed formatting issues

PR #827, opened 2020-08-27
chore: Make server:update command more informative

tolusha:
Signed-off-by: Anatolii Bazko <abazko@redhat.com>

### What does this PR do?
Provides more information to a user before before updating Eclipse Che

https://issues.redhat.com/browse/CRW-1175

### DIFF
--- a/src/commands/server/update.ts
+++ b/src/commands/server/update.ts
@@ -28,7 +28,7 @@ import { PlatformTasks } from '../../tasks/platforms/platform'
 import { getImageTag, isKubernetesPlatformFamily } from '../../util'
 
 export default class Update extends Command {
-  static description = `Update Eclipse Che server to ${getImageTag(DEFAULT_CHE_OPERATOR_IMAGE)} version.`
+  static description = `Update Eclipse Che server to version.`
 
   static flags = {
     installer: string({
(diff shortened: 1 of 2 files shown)


src/commands/server/update.ts
```diff
@@ -25,10 +25,10 @@ import { InstallerTasks } from '../../tasks/installers/installer'
 import { ApiTasks } from '../../tasks/platforms/api'
 import { CommonPlatformTasks } from '../../tasks/platforms/common-platform-tasks'
 import { PlatformTasks } from '../../tasks/platforms/platform'
-import { isKubernetesPlatformFamily } from '../../util'
+import { getImageTag, isKubernetesPlatformFamily } from '../../util'
 
 export default class Update extends Command {
-  static description = 'update Eclipse Che server'
+  static description = `Update Eclipse Che server to ${getImageTag(DEFAULT_CHE_OPERATOR_IMAGE)} version.`
```
mmorhun, line 31: We have hardcoded version of Che server to update to here. So just running `chectl server:update` is not enough (otherwise user will see a message that Che server is up to date, however in reality it could be pretty old). The right flow as of now is to, firstly, update chectl, secondly, Che server. As we cannot just update Che server without chectl update, probably we have to improve the flow. It would be good to check for chectl update if `chectl server:update` says "up to date" and maybe ask user to comfirm automatic update. This could be tricky to implement, so maybe it worth to create a separate issue for that. I think we should at least print warning to user, that it is required to update chectl first using command `****`.

tolusha: I realized that in case of olm the description will be irrelevant.

PR #835, opened 2020-08-31
fix: fix chectl logs before che pods start

flacatus:

### What does this PR do?
Sometimes chectl fails before Che pods start and that means there are not any logs. 
This PR check if directory logs are empty and print a message.

https://issues.redhat.com/browse/CRW-859


flacatus:
image

PR #842, opened 2020-09-03
feat: Rework server:delete to remove consoleLinks, namespace or openshift oauthAuthorizations

flacatus:

### What does this PR do?
This PR have:
 
- New flag(--delete-namespace) for server:delete which allows users to delete the namespace after remove Che resources from cluster
- Remove ConsoleLinks from cluster 
- Remove all oauthClientAuthorizations linked to che instance

https://issues.redhat.com/browse/CRW-714
https://github.com/eclipse/che/issues/17784

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -38,6 +38,7 @@
                 "server:delete",
                 "--chenamespace",
                 "che",
+                "--delete-namespace"
             ],
             "console": "integratedTerminal",
             "internalConsoleOptions": "neverOpen",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -28,7 +28,24 @@
           "windows": {
               "program": "${workspaceFolder}/node_modules/jest/bin/jest",
           }
-      },
+        },
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Debug Openshift server:delete",
+            "program": "${workspaceFolder}/bin/run",
+            "args": [
+                "server:delete",
+                "--chenamespace",
+                "che",
+            ],
+            "console": "integratedTerminal",
+            "internalConsoleOptions": "neverOpen",
+            "disableOptimisticBPs": true,
+            "windows": {
+                "program": "${workspaceFolder}/node_modules/jest/bin/jest",
+            }
+        },
         {
             "type": "node",
             "request": "launch",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -32,24 +32,6 @@
         {
             "type": "node",
             "request": "launch",
-            "name": "Debug Openshift server:delete",
-            "program": "${workspaceFolder}/bin/run",
-            "args": [
-                "server:delete",
-                "--chenamespace",
-                "che",
-                "--delete-namespace"
-            ],
-            "console": "integratedTerminal",
-            "internalConsoleOptions": "neverOpen",
-            "disableOptimisticBPs": true,
-            "windows": {
-                "program": "${workspaceFolder}/node_modules/jest/bin/jest",
-            }
-        },
-        {
-            "type": "node",
-            "request": "launch",
             "name": "Jest All",
             "program": "${workspaceFolder}/node_modules/.bin/jest",
             "args": [
--- a/src/api/kube.ts
+++ b/src/api/kube.ts
@@ -1304,6 +1304,12 @@ export class KubeHelper {
 
       return crs[0]
     } catch (e) {
+      if (e.response && e.response.statusCode === 404) {
+        throw this.wrapK8sClientError('Custom Resource "checlusters.org.eclipse" not found.')
+      } else if (e.response && e.response.statusCode === 403) {
+        throw this.wrapK8sClientError('It looks like you don\'t have enough privileges to delete Custom Resource "checlusters.org.eclipse"')
+      }
+
       throw this.wrapK8sClientError(e)
     }
   }
@@ -1311,12 +1317,16 @@ export class KubeHelper {
   /**
    * Returns all `checlusters.org.eclipse.che' resources
    */
-  async getAllCheCluster(): Promise<any[]> {
+  async getAllCheCluster(): Promise<any[] | undefined> {
     const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
     try {
       const { body } = await customObjectsApi.listClusterCustomObject('org.eclipse.che', 'v1', 'checlusters')
       return body.items ? body.items : []
     } catch (e) {
+      if (e.response.statusCode === 404) {
+        // There is no CRD 'checlusters`
+        return
+      }
       throw this.wrapK8sClientError(e)
     }
   }
@@ -1401,6 +1411,52 @@ export class KubeHelper {
       return this.compare(body, name)
     } catch {
       return false
+    }
+  }
+
+  async getOAuthClientAuthorizations(name: string): Promise<string[]> {
+    const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
+    try {
+      const { body } = await customObjectsApi.listClusterCustomObject('oauth.openshift.io', 'v1', 'oauthclientauthorizations')
+
+      if (!body.items) {
+        return []
+      }
+      const oauthClientAuthorizations = body.items as any[]
+      return oauthClientAuthorizations.filter(o => o.metadata.name && o.clientName === name)
+    } catch {
+      return []
+    }
+  }
+
+  async deleteOAuthClientAuthorizations(oAuthClientAuthorizations: any[]): Promise<boolean> {
+    const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
+    try {
+      const options = new V1DeleteOptions()
+      oAuthClientAuthorizations.forEach(e => customObjectsApi.deleteClusterCustomObject('oauth.openshift.io', 'v1', 'oauthclientauthorizations', e.metadata.name, options))
+      return true
+    } catch (e) {
+      throw this.wrapK8sClientError(e)
+    }
+  }
+
+  async consoleLinkExists(name: string): Promise<boolean> {
+    const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
+    try {
+      const { body } = await customObjectsApi.getNamespacedCustomObject('console.openshift.io', 'v1', '' , 'consolelinks', name)
+      return this.compare(body, name)
+    } catch {
+      return false
+    }
+  }
+
+  async deleteConsoleLink(name: string) {
+    const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
+    try {
+      const options = new V1DeleteOptions()
+      await customObjectsApi.deleteNamespacedCustomObject('console.openshift.io', 'v1', '' , 'consolelinks', name, options)
+    } catch (e) {
+      throw this.wrapK8sClientError(e)
     }
   }
 
--- a/src/api/kube.ts
+++ b/src/api/kube.ts
@@ -1303,8 +1303,12 @@ export class KubeHelper {
       }
 
       return crs[0]
-    } catch {
-      return
+    } catch (e) {
+      if (e.response.statusCode === 404) {
+        // There is no CRD 'checlusters`
+        return
+      }
+      throw this.wrapK8sClientError(e)
     }
   }
 
@@ -1316,8 +1320,12 @@ export class KubeHelper {
     try {
       const { body } = await customObjectsApi.listClusterCustomObject('org.eclipse.che', 'v1', 'checlusters')
       return body.items ? body.items : []
-    } catch {
-      return []
+    } catch (e) {
+      if (e.response.statusCode === 404) {
+        // There is no CRD 'checlusters`
+        return []
+      }
+      throw this.wrapK8sClientError(e)
     }
   }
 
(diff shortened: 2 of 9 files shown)


src/api/kube.ts
```diff
@@ -1304,19 +1304,29 @@ export class KubeHelper {
...
-  async getAllCheCluster(): Promise<any[]> {
+  async getAllCheCluster(): Promise<any[] | undefined> {
     const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
     try {
       const { body } = await customObjectsApi.listClusterCustomObject('org.eclipse.che', 'v1', 'checlusters')
       return body.items ? body.items : []
     } catch (e) {
+      if (e.response.statusCode === 404) {
+        // There is no CRD 'checlusters`
+        return
```
tolusha, line 1328: Just return `[]`

src/api/kube.ts
```diff
@@ -1304,19 +1304,29 @@ export class KubeHelper {
 
       return crs[0]
     } catch (e) {
+      if (e.response && e.response.statusCode === 404) {
+        throw this.wrapK8sClientError('Custom Resource "checlusters.org.eclipse" not found.')
```
tolusha, line 1308: Is it better to `return undefined` ?

src/api/kube.ts
```diff
@@ -1404,6 +1414,52 @@ export class KubeHelper {
     }
   }
 
+  async getOAuthClientAuthorizations(name: string): Promise<string[]> {
```
tolusha, line 1417: `name` -> `clientName`

src/api/kube.ts
```diff
@@ -1404,6 +1414,52 @@ export class KubeHelper {
...
+    const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
+    try {
+      const { body } = await customObjectsApi.listClusterCustomObject('oauth.openshift.io', 'v1', 'oauthclientauthorizations')
+
+      if (!body.items) {
+        return []
+      }
+      const oauthClientAuthorizations = body.items as any[]
+      return oauthClientAuthorizations.filter(o => o.metadata.name && o.clientName === name)
```
tolusha, line 1426: Why do we need to check if `o.metadata.name` exists ?

src/api/kube.ts
```diff
@@ -1404,6 +1414,52 @@ export class KubeHelper {
...
+    } catch (e) {
+      throw this.wrapK8sClientError(e)
+    }
+  }
+
+  async consoleLinkExists(name: string): Promise<boolean> {
+    const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
+    try {
+      const { body } = await customObjectsApi.getNamespacedCustomObject('console.openshift.io', 'v1', '' , 'consolelinks', name)
+      return this.compare(body, name)
```
tolusha, line 1447: We don't need `this.compare` this is kind of legacy.

src/api/kube.ts
```diff
@@ -1404,6 +1414,52 @@ export class KubeHelper {
...
+    } catch {
+      return false
+    }
+  }
+
+  async deleteConsoleLink(name: string) {
+    const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
+    try {
+      const options = new V1DeleteOptions()
+      await customObjectsApi.deleteNamespacedCustomObject('console.openshift.io', 'v1', '' , 'consolelinks', name, options)
```
tolusha, line 1457: Is it namespaced or cluster scope object?

src/commands/server/delete.ts
```diff
@@ -28,6 +28,10 @@ export default class Delete extends Command {
   static flags = {
     help: flags.help({ char: 'h' }),
     chenamespace: cheNamespace,
+    'delete-namespace': boolean({
+      description: 'Set to true to delete namespace after removing Eclipse Che',
```
tolusha, line 32: `Indicates that a Eclipse Che namespace will be deleted as well`

src/api/kube.ts
flacatus: I get the name of `oauthClientAuthorizations` from `o.metadata.name` to delete it.

src/api/kube.ts
flacatus: is cluster scope object... my mistake

src/api/kube.ts
flacatus: Apparently is better to return undefined and don't catch all statusCode.

.vscode/launch.json
```diff
@@ -28,7 +28,25 @@
...
+                "che",
+                "--delete-namespace"
+            ],
+            "console": "integratedTerminal",
+            "internalConsoleOptions": "neverOpen",
+            "disableOptimisticBPs": true,
+            "windows": {
+                "program": "${workspaceFolder}/node_modules/jest/bin/jest",
+            }
+        },
```
mmorhun, line 49: Do we really need it?

src/api/kube.ts
```diff
@@ -1404,6 +1404,52 @@ export class KubeHelper {
...
+    try {
+      const { body } = await customObjectsApi.listClusterCustomObject('oauth.openshift.io', 'v1', 'oauthclientauthorizations')
+
+      if (!body.items) {
+        return []
+      }
+      const oauthClientAuthorizations = body.items as any[]
+      return oauthClientAuthorizations.filter(o => o.metadata.name && o.clientName === clientName)
+    } catch {
+      return []
```
mmorhun, line 1418: Shouldn't we throw an error here?

.vscode/launch.json
flacatus: No, I'll remove it.

src/api/kube.ts
flacatus: Not necessary. We can throw an error in case if for some reason is not possible to obtain `oauthClientAuthorizations`

mmorhun: I meant a situation when `customObjectsApi.listClusterCustomObject` fails due to network or permission or so error. Is it safe to return empty list in such case?

flacatus: yes, it is. This functions will return `[]` . For example if there is some network or permissions the error will be thrown here: https://github.com/flacatus/chectl/blob/delete_cleanup/src/api/kube.ts#L1422-L1431

mmorhun (approved):

src/api/kube.ts
AndrienkoAleksandr: Maybe we should at least log this error.

AndrienkoAleksandr (approved):

src/api/kube.ts
```diff
@@ -1303,8 +1303,8 @@ export class KubeHelper {
       }
 
       return crs[0]
-    } catch (e) {
-      throw this.wrapK8sClientError(e)
+    } catch {
+      return
```
tolusha, line 1307: I think `return undefined` only if we got 404

src/api/kube.ts
```diff
@@ -1316,8 +1316,8 @@ export class KubeHelper {
     try {
       const { body } = await customObjectsApi.listClusterCustomObject('org.eclipse.che', 'v1', 'checlusters')
       return body.items ? body.items : []
-    } catch (e) {
-      throw this.wrapK8sClientError(e)
+    } catch {
+      return []
```
tolusha, line 1320: I think `return []` only if we got 404

src/api/kube.ts
```diff
@@ -1404,6 +1404,52 @@ export class KubeHelper {
...
+  async getOAuthClientAuthorizations(clientName: string): Promise<string[]> {
+    const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
+    try {
+      const { body } = await customObjectsApi.listClusterCustomObject('oauth.openshift.io', 'v1', 'oauthclientauthorizations')
+
+      if (!body.items) {
+        return []
+      }
+      const oauthClientAuthorizations = body.items as any[]
+      return oauthClientAuthorizations.filter(o => o.metadata.name && o.clientName === clientName)
```
tolusha, line 1416: pls don't check `o => o.metadata.name` if we don't need it here.
do this check when it is needed (removing the object)

src/api/kube.ts
```diff
@@ -1404,6 +1404,52 @@ export class KubeHelper {
...
+    }
+  }
+
+  async consoleLinkExists(name: string): Promise<boolean> {
+    const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
+    try {
+      await customObjectsApi.getClusterCustomObject('console.openshift.io', 'v1', 'consolelinks', name)
+      return true
+    } catch {
+      return false
```
tolusha, line 1439: I think `return false` only if we got 404.
Will we have any issues on k8s then?

src/tasks/che.ts
```diff
@@ -461,6 +475,19 @@ export class CheTasks {
...
   }
 
+  deleteNamespace(flags: any): ReadonlyArray<Listr.ListrTask> {
+    return [{
+      title: `Delete namespace ${flags.chenamespace}`,
+      task: async (task: any) => {
+        const namespaceExist = await this.kube.namespaceExist(flags.chenamespace)
+        if (namespaceExist) {
+          await this.kube.deleteNamespace(flags.chenamespace)
+          task.title = await `${task.title}...OK`
```
tolusha, line 485: Anyway print "OK" even if namespace doesn't exist

src/api/kube.ts
flacatus: No, let's return false

src/api/kube.ts
```diff
@@ -1404,6 +1412,64 @@ export class KubeHelper {
...
+      }
+      throw this.wrapK8sClientError(e)
+    }
+  }
+
+  async deleteOAuthClientAuthorizations(oAuthClientAuthorizations: any[]): Promise<boolean> {
+    const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
+    try {
+      const options = new V1DeleteOptions()
+      oAuthClientAuthorizations.forEach(e => {
```
tolusha, line 1438: first `.filter()` then `.forEach`

src/tasks/installers/operator.ts
```diff
@@ -332,6 +332,19 @@ export class OperatorTasks {
...
     return [{
+      title: 'Delete oauthClientAuthorizations',
+      task: async (_ctx: any, task: any) => {
+        const checluster = await kh.getCheCluster(flags.chenamespace)
+        if (checluster && checluster.spec.auth.oAuthClientName) {
+          const oAuthClientAuthorizations = await kh.getOAuthClientAuthorizations(checluster.spec.auth.oAuthClientName)
+          await kh.deleteOAuthClientAuthorizations(oAuthClientAuthorizations)
+          task.title = await `${task.title}...OK`
+        } else {
+          task.title = await `${task.title}...Skipped: No oauthClientAuthorizations found.`
```
tolusha, line 343: Don't print `SKipped` . Just print `OK` anyway.

src/tasks/che.ts
```diff
@@ -461,6 +475,21 @@ export class CheTasks {
...
+        } else {
+          task.title = await `${task.title}...OK`
```
tolusha, line 487: Just print `OK` once after `if` block.

src/api/kube.ts
flacatus: Is better to return undefined because if we return [] or something else https://github.com/che-incubator/chectl/blob/master/src/tasks/installers/operator.ts#L340 will not work properly

src/api/kube.ts
```diff
@@ -1404,6 +1412,63 @@ export class KubeHelper {
...
+      }
+      throw this.wrapK8sClientError(e)
+    }
+  }
+
+  async deleteOAuthClientAuthorizations(oAuthClientAuthorizations: any[]): Promise<boolean> {
+    const customObjectsApi = KubeHelper.KUBE_CONFIG.makeApiClient(CustomObjectsApi)
+    try {
+      const options = new V1DeleteOptions()
+      oAuthClientAuthorizations.filter((e => e.metadata && e.metadata.name)).forEach(async e => {
```
tolusha, line 1438: I hesitates if it really waits until all objects are deleted.

src/tasks/che.ts
```diff
@@ -461,6 +475,20 @@ export class CheTasks {
...
   }
 
+  deleteNamespace(flags: any): ReadonlyArray<Listr.ListrTask> {
+    return [{
+      title: `Delete namespace ${flags.chenamespace}`,
+      task: async (task: any) => {
+        const namespaceExist = await this.kube.namespaceExist(flags.chenamespace)
+        if (namespaceExist) {
+          await this.kube.deleteNamespace(flags.chenamespace)
+          task.title = await `${task.title}...OK`
```
tolusha, line 485: We can remove this line.

src/api/kube.ts
```diff
@@ -1404,6 +1412,68 @@ export class KubeHelper {
...
+    } catch (e) {
+      if (e.response.statusCode === 404) {
+        // There is no 'oauthclientauthorizations`
+        return []
+      }
+      throw this.wrapK8sClientError(e)
+    }
+  }
+
+  async deleteOAuthClientAuthorizations(oAuthClientAuthorizations: any[]): Promise<boolean> {
```
tolusha, line 1434: Just return void instead of boolean

src/api/kube.ts
```diff
@@ -1404,6 +1412,68 @@ export class KubeHelper {
...
+    } catch (e) {
+      if (e.response.statusCode === 404) {
+        // There are no consoleLink
+        return false
+      }
+      throw this.wrapK8sClientError(e)
+    }
+  }
+
+  async deleteConsoleLink(name: string) {
```
tolusha, line 1467: add :void

tolusha (approved):

flacatus:
I've finished to test new changes in minikube and openshift and seems everything is correct:

Openshift: 
1. Install Eclipse Che with OAuth +  login into Eclipse Che [I create 2 users to create oauthAuthorizations]
2. server:delete, check if any resources exist(consoleLink, oauthAuths...)... work fine
3. Execute again and check if fix Angel issue [Object object] (https://github.com/eclipse/che/issues/17784).. Fixed
4. Install Eclipse Che without OAuth
5. server:delete to check if resources are removed... Work fine.

Minikube:
Basically I did the same steps like for openshift but without oauth and seems everything works fine

flacatus:
[test]

PR #849, opened 2020-09-09
feat: Rework workspace commands to be able to work with user namespace permissions only

mmorhun:

### What does this PR do?
All `workspace` commands are reworked in order to be able to work correctly if current user doesn't have permissions to read Che Server namespace. However, in such case we cannot automatically retrieve Che API URL, so it is required from user to provide it in `--che-api-endpoint` argument.
Also this PR moves all Che API related methods into a separate singleton class.

### How to test

The following commands are supposed to be run manually one by one on a CRC cluster with a user which has no permissions for Che Server namespace.

<details>

<summary>Test commands</summary>

```bash
OS_TOKEN=$(oc whoami --show-token)
AT=$(curl -k -X POST -d "client_id=che-public" -d "subject_token=${OS_TOKEN}" -d "subject_issuer=openshift-v4" --data-urlencode "grant_type=urn:ietf:params:oauth:grant-type:token-exchange" --data-urlencode "subject_token_type=urn:ietf:params:oauth:token-type:access_token" https://keycloak-che.apps-crc.testing/auth/realms/che/protocol/openid-connect/token | jq -r .access_token)
ARGS="--access-token=$AT --che-api-endpoint=https://che-che.apps-crc.testing/api"

cat <<EOF >> devfile.yaml
metadata:
  generateName: wksp-custom-
apiVersion: 1.0.0
EOF

./chectl workspace:create $ARGS
./chectl workspace:list $ARGS
./chectl workspace:start $WID $ARGS
./chectl workspace:logs --workspace $WID --namespace=user-che
./chectl workspace:inject -k $ARGS
./chectl workspace:stop $WID $ARGS
./chectl workspace:delete $WID $ARGS
```

</details>

https://issues.redhat.com/browse/CRW-803


AndrienkoAleksandr:
@mmorhun you have failed tests.

src/api/che-api.ts
```diff
@@ -0,0 +1,340 @@
...
+    } catch {
+      throw new Error(`E_CHE_API_URL_NO_RESPONSE - Failed to connect to "${this.cheApiUrl}". Is it the right url?`)
+    }
+    throw new Error(`E_CHE_API_WRONG_URL - Provided url "${this.cheApiUrl}" is not Che API url`)
+  }
+
+  async isCheServerReady(responseTimeoutMs = this.defaultCheResponseTimeoutMs): Promise<boolean> {
+    const id = await this.axios.interceptors.response.use(response => response, async (error: any) => {
+      if (error.config && error.response && (error.response.status === 404 || error.response.status === 503)) {
+        return this.axios.request(error.config)
```
AndrienkoAleksandr, line 78: I guess this code made requests while Che will be ready. Do we need timeout for this interceptor recursion?

mmorhun: Yes, good catch, thanks.

flacatus:
[test]

flacatus:
@mmorhun Openshift CI seems today is unstable. Ci centos jobs are removing so, should not work in this PR. Should be run tests in github actions: https://github.com/che-incubator/chectl/pull/851

src/api/kube.ts
```diff
@@ -555,6 +555,20 @@ export class KubeHelper {
     }
   }
 
+  async hasReadPermissionsForNamespace(namespace: string): Promise<boolean> {
```
tolusha, line 558: `hasNamespaceReadPermissions`

src/api/kube.ts
```diff
@@ -1646,11 +1660,17 @@ export class KubeHelper {
...
-  async deleteNamespace(namespace: string): Promise<void> {
+  async deleteNamespace(namespace: string): Promise<boolean> {
     const k8sCoreApi = KubeHelper.KUBE_CONFIG.makeApiClient(CoreV1Api)
     try {
       await k8sCoreApi.deleteNamespace(namespace)
+      return true
     } catch (e) {
+      if (e.response && e.response.body) {
+        if (e.response.body.code === 403) {
+          return false
```
tolusha, line 1671: That's a bit strange. The Namespace isn't deleted in both cases but only for 403 we return false.
It seems inconsistent for me.

src/common-flags.ts
```diff
@@ -39,3 +40,10 @@ export const skipKubeHealthzCheck = boolean({
   description: 'Skip Kubernetes health check',
   default: false
 })
+
+export const CHE_API_URL_KEY = 'che-api-url'
+export const cheApiUrl = string({
+  description: 'Eclipse Che server API URL',
```
tolusha, line 46: URL -> endpoint

src/common-flags.ts
```diff
@@ -39,3 +40,10 @@ export const skipKubeHealthzCheck = boolean({
...
+  env: 'CHE_API_URL',
```
tolusha, line 47: `CHE_API_ENDPOINT`

src/commands/workspace/stop.ts
```diff
@@ -39,28 +38,26 @@ export default class Stop extends Command {
...
+    const workspaceId = args.workspace
 
-    const apiTasks = new ApiTasks()
-    const cheTasks = new CheTasks(flags)
-    const workspaceTasks = new WorkspaceTasks(flags)
+    let cheApiUrl = flags[CHE_API_URL_KEY]
+    if (!cheApiUrl) {
+      const kube = new KubeHelper(flags)
+      if (!await kube.hasReadPermissionsForNamespace(flags.chenamespace)) {
+        throw new Error(`"--${CHE_API_URL_KEY}" argument is required`)
```
tolusha, line 48: `Use flag --${CHE_API_URL_KEY} to provide Eclipse Che API endpoint`
pls update all errors like this

src/commands/workspace/stop.ts
```diff
@@ -10,22 +10,21 @@
...
 export default class Stop extends Command {
   static description = 'Stop a running workspace'
 
   static flags = {
     help: flags.help({ char: 'h' }),
-    'access-token': accessToken,
+    [CHE_API_URL_KEY]: cheApiUrl,
+    [ACCESS_TOKEN_KEY]: accessToken,
     chenamespace: cheNamespace,
-    'skip-kubernetes-health-check': skipKubeHealthzCheck
```
tolusha, line 28: Pls, don't remove this flag even it isn't used. 
The commands the use it will fail. 
Restore it in all commands.

src/commands/workspace/list.ts
```diff
@@ -10,54 +10,55 @@
...
+import { CheApiClient } from '../../api/che-api-client'
+import { KubeHelper } from '../../api/kube'
+import { accessToken, ACCESS_TOKEN_KEY, cheApiUrl, cheNamespace, CHE_API_URL_KEY } from '../../common-flags'
 
 export default class List extends Command {
   static description = 'list workspaces'
 
   static flags = {
     help: flags.help({ char: 'h' }),
+    quiet: flags.boolean({
```
tolusha, line 24: Why do we need this?
Pls remove. User always can use bash script to extract workspaceId from the output.

mmorhun: I think it is usable. For end user it will be much easier to parse a list of ID's instead of a table of values.

src/api/che-api-client.ts
```diff
@@ -0,0 +1,343 @@
...
+  async getAllWorkspaces(accessToken?: string): Promise<chetypes.workspace.Workspace[]> {
+    const all: any[] = []
+    const maxItems = 30
+    let skipCount = 0
+
+    do {
+      const workspaces = await this.getWorkspaces(skipCount, maxItems, accessToken)
+      all.push(...workspaces)
+      skipCount += workspaces.length
+    } while (all.length === maxItems)
```
AndrienkoAleksandr, line 152: I think here should be `} while (workspaces.length === maxItems)` to retrieve really all workspaces... Current logic could retrieve no more than 60 workspaces....

AndrienkoAleksandr, line 152: Here should be something like that:

```ts
    let length
    do {
      const workspaces = await this.getWorkspaces(skipCount, maxItems, accessToken)
      all.push(...workspaces)
      skipCount += workspaces.length
      length = workspaces.length
    } while (all.length === maxItems)
```

Current logic can't retrieve more than 60 workspaces.

src/api/che-api-client.ts
```diff
@@ -0,0 +1,343 @@
...
+ * Singleton responsible for calls to Che API.
+ */
+let instance: CheApiClient | undefined
+export class CheApiClient {
+  public defaultCheResponseTimeoutMs = 3000
+  public readonly cheApiUrl: string
+
+  private readonly axios: AxiosInstance
+
+  private constructor(cheApiUrl: string) {
```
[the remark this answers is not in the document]
mmorhun: Ok

src/api/che-api-client.ts
```diff
@@ -0,0 +1,343 @@
...
+      }
+      return url
+    }
+  }
+
+  /**
+   * Checks whether provided url really points to Che server API.
+   * Trows an exception if not.
+   */
+  async ensureCheApiUrlCorrect(responseTimeoutMs = this.defaultCheResponseTimeoutMs): Promise<void> {
```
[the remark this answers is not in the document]
mmorhun: 1) I am not sure... If we needed to set it only once, then the answer would be yes. But here we need some flexibility and should be able to override the option.
2) I think `check` will be better
3) No. This is singleton object. We create it once and then might get it in other places many times, but the check should be done only once.

src/api/che.ts
```diff
@@ -205,166 +205,55 @@ export class CheHelper {
     return this.kube.namespaceExist(namespace)
   }
 
+  /**
+   * DEPRECATED. Use CheApiClient instead.
+   */
```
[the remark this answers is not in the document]
mmorhun: Yes, we unfortunately do, at least now.
I moved methods related to Che server API from the file. But there is still some Che related stuff in it (however, I think, many of them, should be moved into kubernetes helper, let's leave it for one of the next PRs).

src/api/kube.ts
mmorhun: To me current version is better...

src/api/kube.ts
mmorhun: True. Will fix.

mmorhun: Agree, will revert boolean return change.

src/commands/workspace/stop.ts
mmorhun: I think this flag is used really rarely... @tolusha do you insist on restoring it with deprecation message?

src/api/che-api-client.ts
```diff
@@ -0,0 +1,343 @@
...
+      }
+    }
+
+    if (!response || response.status !== 200 || !response.data) {
+      throw new Error('E_BAD_RESP_CHE_API')
+    }
+  }
+
+  async stopWorkspace(workspaceId: string, accessToken?: string): Promise<void> {
+    let endpoint = `${this.cheApiUrl}/workspace/${workspaceId}/runtime`
```
AndrienkoAleksandr, line 243: let -> const ?

AndrienkoAleksandr: And  Should be the same in the other places?

src/api/che-api-client.ts
mmorhun: Which will probably fail too. I think it should be like:
```typescript
    let skipCount = 0
    let workspaces = []
    do {
      workspaces = await this.getWorkspaces(skipCount, itemsPerPage, accessToken)
      all.push(...workspaces)
      skipCount += workspaces.length
    } while (workspaces.length === itemsPerPage)
```

src/commands/workspace/list.ts
tolusha: Pls remove it.

src/commands/workspace/stop.ts
tolusha: No, just keep the flag.

src/api/che.ts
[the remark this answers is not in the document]
tolusha: ok

src/commands/workspace/list.ts
mmorhun: Okay =(

src/api/kube.ts
AndrienkoAleksandr, line 558: k8s provides for us api to access estimation. Could you use it? :

Proposed changes:

```typescript
    const k8sApi = KubeHelper.KUBE_CONFIG.makeApiClient(AuthorizationV1Api)
    const review = new V1SelfSubjectAccessReview()
    review.spec = {}
    review.spec.resourceAttributes = {
      group: "", 
      name: "che-access-to-namespace", 
      namespace, 
      resource: "namespaces", 
      verb: "get"
    }

    try {
      const { body } = await k8sApi.createSelfSubjectAccessReview(review)
      return body.status!.allowed
    } catch (error) {
      if (error.response && error.response.body) {
        if (error.response.body.code === 403) {
          return false
        }
      }
      throw this.wrapK8sClientError(error)
    }

```
For me it looks very similar to this one pr: https://github.com/eclipse/che-operator/pull/209

AndrienkoAleksandr (approved):

mmorhun:
Rebased

tolusha (approved):

PR #870, opened 2020-09-22
feat: Create new chectl command -> server:status

flacatus:

### What does this PR do?
This PR add a new command to chectl which prints some information about che instance: if Openshift Oauth is enabled, version, che url or how many workspaces exist.
Also this PR rename the variable of minishift e2e job (GITHUBTOKEN--> GITHUB_TOKEN)

https://github.com/eclipse/che/issues/14476

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ USAGE
 * [`chectl server:delete`](#chectl-serverdelete)
 * [`chectl server:logs`](#chectl-serverlogs)
 * [`chectl server:start`](#chectl-serverstart)
+* [`chectl server:status`](#chectl-serverstatus)
 * [`chectl server:stop`](#chectl-serverstop)
 * [`chectl server:update`](#chectl-serverupdate)
 * [`chectl update [CHANNEL]`](#chectl-update-channel)
@@ -445,6 +446,27 @@ OPTIONS
 ```
 
 _See code: [src/commands/server/start.ts](https://github.com/che-incubator/chectl/blob/v0.0.2/src/commands/server/start.ts)_
+
+## `chectl server:status`
+
+status Eclipse Che server
+
+```
+USAGE
+  $ chectl server:status
+
+OPTIONS
+  -h, --help                               show CLI help
+
+  -n, --chenamespace=chenamespace          [default: che] Kubernetes namespace where Eclipse Che server is supposed to
+                                           be deployed
+
+  --listr-renderer=default|silent|verbose  [default: default] Listr renderer
+
+  --skip-kubernetes-health-check           Skip Kubernetes health check
+```
+
+_See code: [src/commands/server/status.ts](https://github.com/che-incubator/chectl/blob/v0.0.2/src/commands/server/status.ts)_
 
 ## `chectl server:stop`
 
--- a/README.md
+++ b/README.md
@@ -456,14 +456,10 @@ USAGE
   $ chectl server:status
 
 OPTIONS
-  -h, --help                               show CLI help
+  -h, --help                       show CLI help
 
-  -n, --chenamespace=chenamespace          [default: che] Kubernetes namespace where Eclipse Che server is supposed to
-                                           be deployed
-
-  --listr-renderer=default|silent|verbose  [default: default] Listr renderer
-
-  --skip-kubernetes-health-check           Skip Kubernetes health check
+  -n, --chenamespace=chenamespace  [default: che] Kubernetes namespace where Eclipse Che server is supposed to be
+                                   deployed
 ```
 
 _See code: [src/commands/server/status.ts](https://github.com/che-incubator/chectl/blob/v0.0.2/src/commands/server/status.ts)_
--- a/src/api/version.ts
+++ b/src/api/version.ts
@@ -20,6 +20,7 @@ export namespace VersionHelper {
   export const MINIMAL_K8S_VERSION = '1.9'
   export const MINIMAL_HELM_VERSION = '2.15'
   export const CHE_POD_MANIFEST_FILE = '/home/user/eclipse-che/tomcat/webapps/ROOT/META-INF/MANIFEST.MF'
+  export const CHE_PREFFIX_VERSION = 'Implementation-Version: '
 
   export function getOpenShiftCheckVersionTask(flags: any): Listr.ListrTask {
     return {
@@ -135,12 +136,12 @@ export namespace VersionHelper {
     return stdout.split('\n').filter(value => value.startsWith(versionPrefix)).map(value => value.substring(versionPrefix.length))[0]
   }
 
-  export async function getCheVersionFromPod(namespace: string, podName: string, versionPrefix: string): Promise<string> {
+  export async function getCheVersionFromPod(namespace: string, podName: string): Promise<string> {
     const command = getClusterClientCommand()
     const args = ['exec', podName, '--namespace', namespace, 'cat', CHE_POD_MANIFEST_FILE]
     try {
       const { stdout } = await execa(command, args, { timeout: 60000 })
-      return stdout.split('\n').filter(value => value.startsWith(versionPrefix)).map(value => value.substring(versionPrefix.length))[0]
+      return stdout.split('\n').filter(value => value.startsWith(CHE_PREFFIX_VERSION)).map(value => value.substring(CHE_PREFFIX_VERSION.length))[0]
     } catch (error) {
       throw new Error(`E_COMMAND_FAILED: ${error}`)
     }
--- a/src/api/version.ts
+++ b/src/api/version.ts
@@ -11,6 +11,8 @@
 import execa = require('execa')
 import Listr = require('listr')
 
+import { getClusterClientCommand } from '../util'
+
 import { KubeHelper } from './kube'
 
 export namespace VersionHelper {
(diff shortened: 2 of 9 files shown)


flacatus:
Hi. About print the info aligned in bash terminal I found only this way: https://github.com/che-incubator/chectl/pull/870/files#diff-460d15365f3c03a6a088ecdc731b0b0eR66-R69. If someone know a better approach, Will be a pleasure to add it in my PR.

flacatus:
The current output it is:
image

AndrienkoAleksandr:
> Hi. About print the info aligned in bash terminal I found only this way: https://github.com/che-incubator/chectl/pull/870/files#diff-460d15365f3c03a6a088ecdc731b0b0eR66-R69. If someone know a better approach, Will be a pleasure to add it in my PR.

You can try to use cli-table from https://github.com/oclif/cli-ux
https://github.com/che-incubator/chectl/blob/master/src/commands/workspace/list.ts#L65

flacatus:
I've tried cli.table but prints only in columns from what I see in https://github.com/oclif/cli-ux#clitable

AndrienkoAleksandr:
```
const columns: Table.Columns = {
  // where `.name` is a property of a data object
  name: {}, // "Name" inferred as the column header
  id: {
    header: 'ID', // override column header
    minWidth: '10', // column must display at this width or greater
    extended: true, // only display this column when the --extended flag is present
    get: row => `US-O1-${row.id}`, // custom getter for data row object
  },
}
```

Maybe you should try to use:
get: row => `US-O1-${row.id}`, // custom getter for data row object

flacatus:
I've tried that as well but the output still be:
image

tolusha:
In general the only flag is accepted is `-n`.
This command is supposed to be run by user who has cluster admin rights.
1. Is it possible to get che version without making request? if no, let's print che-server image
2.`OpenShift OAuth enabled` we have to analyze CR (if exists) to check if `server.auth.openShiftoAuth` is set and print yes/no

flacatus:
@tolusha It is possible to get che version from CR... In case of nightly deployments(7.19.0-SNAPSHOT) will print only nightly.

tolusha:
inside che-server pod:
`cat /home/user/eclipse-che/tomcat/webapps/ROOT/META-INF/MANIFEST.MF`
```
...
Implementation-Vendor: Red Hat, Inc.
Implementation-Version: 7.19.0-SNAPSHOT
...
```

tolusha:
without access-token it is not possible to get list of workspaces, so let's remove this info.
`workspace:list` command exists for this case.

src/commands/server/status.ts
```diff
@@ -0,0 +1,75 @@
+/*********************************************************************
+ * Copyright (c) 2019 Red Hat, Inc.
```
mmorhun, line 2: Wrong year

src/commands/server/status.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export default class List extends Command {
+  static description = 'list workspaces'
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    chenamespace: cheNamespace,
+    [CHE_API_ENDPOINT_KEY]: cheApiEndpoint,
+    [ACCESS_TOKEN_KEY]: accessToken,
+    'skip-kubernetes-health-check': skipKubeHealthzCheck
```
mmorhun, line 28: It's not needed, please remove.

mmorhun (approved):

src/api/version.ts
```diff
@@ -132,6 +133,20 @@ export namespace VersionHelper {
     return stdout.split('\n').filter(value => value.startsWith(versionPrefix)).map(value => value.substring(versionPrefix.length))[0]
   }
 
+  export async function getCheVersionWithKubectl(namespace: string, podName: string, versionPrefix: string): Promise<string | undefined> {
```
tolusha, line 136: It is not needed to separate into two functions depending on client
 https://github.com/che-incubator/chectl/blob/master/src/util.ts#L18

README.md
```diff
@@ -446,6 +447,27 @@ OPTIONS
...
+USAGE
+  $ chectl server:status
+
+OPTIONS
+  -h, --help                               show CLI help
+
+  -n, --chenamespace=chenamespace          [default: che] Kubernetes namespace where Eclipse Che server is supposed to
+                                           be deployed
+
+  --listr-renderer=default|silent|verbose  [default: default] Listr renderer
```
tolusha, line 464: pls. update README.md (yarn pack)

src/commands/server/status.ts
```diff
@@ -0,0 +1,74 @@
...
+  async run() {
+    const { flags } = this.parse(List)
+    const cr = await this.kube.getCheCluster(flags.chenamespace)
+
+    if (cr && cr.spec && cr.spec.auth && typeof cr.spec.auth.openShiftoAuth === 'boolean') {
+      this.openshiftOauth = true
+    }
+
+    if (cr && cr.status && cr.status.cheURL) {
+      this.cheUrl = cr.status.cheURL
```
tolusha, line 45: The problem is that CR might not exists if `helm` was used.
So, grab the cheURL from the `che` configmap

src/commands/server/status.ts
```diff
@@ -0,0 +1,74 @@
...
+    help: flags.help({ char: 'h' }),
+    chenamespace: cheNamespace,
+  }
+
+  async run() {
+    const { flags } = this.parse(List)
+    const cr = await this.kube.getCheCluster(flags.chenamespace)
+
+    if (cr && cr.spec && cr.spec.auth && typeof cr.spec.auth.openShiftoAuth === 'boolean') {
+      this.openshiftOauth = true
```
tolusha, line 41: The problem is that CR might not exists if helm was used.

src/commands/server/status.ts
```diff
@@ -0,0 +1,74 @@
...
+import * as notifier from 'node-notifier'
+
+import { KubeHelper } from '../../api/kube'
+import { VersionHelper } from '../../api/version'
+import { cheNamespace } from '../../common-flags'
+
+export default class List extends Command {
+  // Implementation-Version it is a property from Manifest.ml inside of che server pod which indicate Eclipse Che build version.
+  readonly chePreffixVersion = 'Implementation-Version: '
+  readonly cheServerSelector = 'app=che,component=che'
```
tolusha, line 22: It is already there https://github.com/che-incubator/chectl/blob/master/src/tasks/che.ts#L33

src/commands/server/status.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  static description = 'status Eclipse Che server'
+  openshiftOauth = false
```
tolusha, line 25: It is better to print Yes/No instead of True/False

src/commands/server/status.ts
flacatus: @tolusha I've reused https://github.com/che-incubator/chectl/blob/master/src/api/che.ts#L102 instead of get url from configMap

src/commands/server/status.ts
```diff
@@ -0,0 +1,60 @@
...
+import * as notifier from 'node-notifier'
+
+import { CheHelper } from '../../api/che'
+import { KubeHelper } from '../../api/kube'
+import { VersionHelper } from '../../api/version'
+import { cheNamespace } from '../../common-flags'
+import { CheTasks } from '../../tasks/che'
+export default class List extends Command {
+  // Implementation-Version it is a property from Manifest.ml inside of che server pod which indicate Eclipse Che build version.
+  readonly chePreffixVersion = 'Implementation-Version: '
```
tolusha, line 22: pls. move this const to `version.ts`
it is not used here.

src/commands/server/status.ts
```diff
@@ -0,0 +1,60 @@
...
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    chenamespace: cheNamespace,
+  }
+
+  async run() {
+    const { flags } = this.parse(List)
+    const cr = await this.kube.getCheCluster(flags.chenamespace)
+
+    if (cr && cr.spec && cr.spec.auth && typeof cr.spec.auth.openShiftoAuth === 'boolean') {
```
tolusha, line 40: `cr.spec.auth.openShiftoAuth` might exist but set to `false`

src/commands/server/status.ts
```diff
@@ -0,0 +1,60 @@
...
+import { cheNamespace } from '../../common-flags'
+import { CheTasks } from '../../tasks/che'
+export default class List extends Command {
+  // Implementation-Version it is a property from Manifest.ml inside of che server pod which indicate Eclipse Che build version.
+  readonly chePreffixVersion = 'Implementation-Version: '
+  static description = 'status Eclipse Che server'
+  openshiftOauth = 'No'
+  cheVersion = 'UNKNOWN'
+
+  kube = new KubeHelper(flags)
```
tolusha, line 27: I am not sure if initialization flow is correct. 
First  `const { flags } = this.parse(List)` then `new ..Helper(flags)`
In your case `KubeHelper` is initialized with `static flags`

src/commands/server/status.ts
```diff
@@ -0,0 +1,60 @@
...
+import { CheHelper } from '../../api/che'
+import { KubeHelper } from '../../api/kube'
+import { VersionHelper } from '../../api/version'
+import { cheNamespace } from '../../common-flags'
+import { CheTasks } from '../../tasks/che'
+export default class List extends Command {
+  // Implementation-Version it is a property from Manifest.ml inside of che server pod which indicate Eclipse Che build version.
+  readonly chePreffixVersion = 'Implementation-Version: '
+  static description = 'status Eclipse Che server'
+  openshiftOauth = 'No'
```
tolusha, line 24: `openshiftOauth` and `cheVersion` should be moved to local variables in `run` function .
They are useless as global ones.

tolusha (approved):

PR #1020, opened 2020-12-03
Release version 7.22.2

flacatus:


nickboldt:
closed, will regen once more using Flavius' GH credentials instead of Mykhailo's

nickboldt:
see https://github.com/che-incubator/chectl/runs/1494507448?check_suite_focus=true

PR #1079, opened 2021-01-29
Release version 7.25.1

flacatus:


mkuznyetsov:
happy path check will be ignored due to infra problems

flacatus:
> can happy path check be ignored?

yes, the fail for happy path tests are not related with the PR.

PR #1182, opened 2021-04-16
Release version 7.29.0

flacatus:


flacatus:
@tolusha @mkuznyetsov. We should skip happy path tests because the original github actions[https://github.com/che-incubator/happy-path-tests-action] is using latest minikube version and che have an known issue: https://github.com/eclipse/che/issues/19602

